PR #7, opened 2019-03-22
feat: added theme

shmidt-i:


src/theme/theme.ts
```diff
@@ -0,0 +1,81 @@
...
+        assert(sizeValues.includes(s1), `[theme.media] Unrecognized size value: ${s1}`);
+        assert(sizeValues.includes(s2), `[theme.media] Unrecognized size value: ${s2}`);
+
+        return `@media (min-width: ${s1}px) and (max-width: ${s2}px)`;
+      },
+      greaterThan: s => {
+        assert(sizeValues.includes(s), `[theme.media] Unrecognized size value: ${size}`);
+        return `@media (min-width: ${size}px)`;
+      },
+      lessThan: s => {
```
mjmonline: I'm really against removing this function, we should not

PR #24, opened 2019-03-27
feat: cards

mjmonline:


src/Atoms/Card/Card.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+  align-items: center;
+  justify-content: center;
+`;
+
+export const Content: React.FC<Props> = ({ children }) => <StyledContent>{children}</StyledContent>;
+
+stories.add('Basic Card', () => {
+  return (
+    <Card>
+      <Content>A Card as a div containing content</Content>
```
JacobBlomgren: maybe use `Text`?

src/theme/theme.types.ts
```diff
@@ -65,8 +65,10 @@ export type Theme = {
     negative: RawColor['negative'];
     /** index */
     warning: RawColor['index'];
-    /** white */
+    /** @deprecated use card instead */
```
JacobBlomgren: since we haven't released yet you don't need to deprecate, you can just remove. Or what do you think @shmidt-i?

src/Atoms/Card/Card.stories.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import styled from 'styled-components';
+import Card from '.';
+import { Props } from './Card.types';
+
+const stories = storiesOf('Atoms/Card', module);
+
+const StyledContent = styled.div`
+  height: 50vh;
```
Maxenboy: should the card always be 50% of screen?

src/theme/theme.types.ts
shmidt-i: We wanted to try out the deprecation process. Also, since guys have started their work, we can be slightly out of sync, then again, deprecation is better

PR #26, opened 2019-03-27
feat: development

kvasdopil:
Initial implementation for development component

Screenshot 2019-03-27 at 16 26 07

### DIFF
--- a/src/Atoms/Development/Development.tsx
+++ b/src/Atoms/Development/Development.tsx
@@ -1,0 +1,16 @@
+import React from 'react';
+import { DevelopmentComponent } from './Development.types';
+import Text from '../Text/index';
+
+const Primary: DevelopmentComponent['Primary'] = ({ value, ...rest }) => (
+  <Text.Primary {...rest}>{`${value}`}</Text.Primary>
+);
+
+const Secondary: DevelopmentComponent['Secondary'] = ({ value, ...rest }) => (
+  <Text.Secondary {...rest}>{`${value}`}</Text.Secondary>
+);
+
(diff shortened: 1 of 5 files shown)


src/Atoms/Text/Text.types.ts
```diff
@@ -8,7 +8,8 @@ export type StyledTextProps = {
   colorFn?: ColorFn;
   weight: Weight;
 };
-type ColorFn = (t: Theme) => keyof Theme['color'];
+
+type ColorFn = (t: Theme) => Theme['color'][keyof Theme['color']];
```
shmidt-i: type Values<ObjectType> = ObjectType extends Record<any, infer K> ? K : never; // can move it to util types
type ColorFn = (t: Theme) => Values<Theme['color']>

src/Atoms/Development/Development.tsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import { DevelopmentComponent } from './Development.types';
+import Text from '../Text/index';
+import { Theme } from '../../theme';
+
+const getPrefix = (value: Number) => {
+  if (value > 0) {
+    return `▲ +${value}%`;
```
mjmonline: You don't want to print the triangle like that. When I run the screen reader it says "Up pointing triangle +12334%".
We either do the triangle with CSS `:before` or put it in a span that's hidden for screen readers

JacobBlomgren: This is a very MVP version so we'll make sure to fix it later

PR #33, opened 2019-03-29
Feat/list

mjmonline:


src/Atoms/List/List.types.ts
```diff
@@ -0,0 +1,4 @@
+export type Props = {
+  children: React.ReactNode;
+  as?: 'ul' | 'ol';
```
shmidt-i: className? or how we gonna use this?

src/Atoms/ListItem/ListItem.types.ts
```diff
@@ -0,0 +1,3 @@
+export type Props = {
+  children: React.ReactNode;
```
shmidt-i: className here too probably

PR #41, opened 2019-04-02
Fix exports

kvasdopil:
Fixing exports for components. From now on all .story files should import components they use from root src folder i.e. `import { Something } from '../..';`. 

Placeholder component is not imported because its export name is probably incorrect and needs to be fixed in another PR.


src/__snapshots__/storyshots.test.ts.snap
```diff
@@ -3487,7 +3487,7 @@ exports[`Storyshots Atoms/Development Tertiary 1`] = `
         className="Text__StyledText-sc-1974afi-0 eumSjM"
       >
         ▼
-        -2.1
+        −2,1
```
shmidt-i: Is this changing every time ? 😄

PR #42, opened 2019-04-02
feat: icon base

mjmonline:
DND-28


mjmonline:
Should we limit the size prop somehow?

src/Atoms/Icon/Icon.stories.tsx
```diff
@@ -1,13 +1,39 @@
...
+import styled from 'styled-components';
 import Icon from '.';
 
 const stories = storiesOf('Atoms/Icon', module);
 
+const StyledContainer = styled.div`
+  display: 'flex';
+  flex-wrap: wrap;
+  margin: 0 -${p => p.theme.spacing.unit(2)}px;
+  & > svg {
```
shmidt-i: Seems to be too much into the implementation details from the stories perspective. Can we remove it?

src/Atoms/Icon/Icon.stories.tsx
```diff
@@ -1,13 +1,39 @@
...
+
+stories.add('Icons with different size and color', () => {
+  return (
+    <Container>
+      <Icon.QuestionMark size={10} color={t => t.color.positive} />
+      <Icon.ChevronUp size={10} color={t => t.color.positive} />
+      <Icon.ChevronDown size={10} color={t => t.color.positive} />
+      <Icon.ArrowRight size={10} color={t => t.color.positive} />
+    </Container>
   );
```
shmidt-i: We need to have extra story of every icon presented

kvasdopil: Do we? IMO single story for all icons should be enough, also it gives good overview on options. Only thing i'd add is icon names to be displayed along with icons.

shmidt-i: Yea, that's what I meant, got confused by the look in the storybook, since the all are houses for now :)

mjmonline: @shmidt-i @kvasdopil check latest commit

src/Atoms/Icon/Icon.stories.tsx
mjmonline: solved in latest commit @shmidt-i

src/Atoms/Icon/components/ChevronUp.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { IconBase } from '../IconBase';
+import { BaseProps } from '../IconBase.types';
+
+export const ChevronUp = (props: BaseProps) => {
+  return (
+    <IconBase {...props}>
+        <polygon points="19.969 4 24 8.027 12 20 0 8.027 4.031 4 12 11.952" transform="rotate(-180 12 12)"/>
```
JacobBlomgren: maybe it could reuse ChevronDown? We're gonna get a lot of those eventually, and the icons are always the heaviest part

PR #44, opened 2019-04-03
feat: normalize

mjmonline:
The small amount of actual elements targeted by our trimmed normalize can live in a global css in my opinion

### DIFF
--- a/src/Molecules/Button/Button.tsx
+++ b/src/Molecules/Button/Button.tsx
@@ -50,7 +50,49 @@ const getHeight = (props: ThemedStyledProps<Props, Theme>) => {
   return `${theme.spacing.unit(hugeness)}px`;
 };
 
-const StyledButton = styled.button<Props>`
+const StyledNeutralButton = styled.button`
+  /**
+   * From Normalize.css v8.0.1
+   * 1. Change the font styles in all browsers.
+   * 2. Remove the margin in Firefox and Safari.
+   * 3. Show the overflow in Edge.
+   * 4. Remove the inheritance of text transform in Edge, Firefox, and IE
+   * 5. Correct the inability to style clickable types in iOS and Safari.
+   * 6. Remove the inner border and padding in Firefox.
+   * 7. Restore the focus styles unset by the previous rule
+   */
+
+  font-family: inherit; /* 1 */
+  font-size: 100%; /* 1 */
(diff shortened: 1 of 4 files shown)


mjmonline:
@iamstarkov Looking at the CSS in this PR it's still bloated, we really only need to normalize `<button>`, `<input>`, `<select>` and `<textarea>`. And the normalize of these elements should live inside the core components themselves

iamstarkov:
hmm 🤔, now its just a button styles

src/Molecules/Button/Button.tsx
```diff
@@ -50,7 +50,39 @@ const getHeight = (props: ThemedStyledProps<Props, Theme>) => {
...
+  line-height: 1.15;
+  margin: 0;
+  overflow: visible;
+  text-transform: none;
+
+  &,
+  &[type='button'],
+  &[type='reset'],
+  &[type='submit'] {
+    -webkit-appearance: button; /* stylelint-disable-line */
```
JacobBlomgren: just a noob question, what does styled components do with already prefixed stuff, like this? might it be stripped out?

src/Molecules/Button/Button.tsx
```diff
@@ -50,7 +50,39 @@ const getHeight = (props: ThemedStyledProps<Props, Theme>) => {
   return `${theme.spacing.unit(hugeness)}px`;
 };
 
-const StyledButton = styled.button<Props>`
+/** From Normalize.css v8.0.1 */
+const NormalizedButton = styled.button`
```
JacobBlomgren: Could this be moved to utils? I have `Tab` component right now in `Tabs`, that probably can't and shouldn't reuse `Button` (though up for discussion), and I would like to reuse this then.

src/Molecules/Button/Button.tsx
mjmonline: This should be ignored and printed out as is

mjmonline: > hmm 🤔, now its just a button styles

Using individual resets within every component seems more in harmony with the styled-components philosophy to me

src/Molecules/Button/Button.tsx
mjmonline: Moved to comon folder

iamstarkov:
👍

PR #53, opened 2019-04-08
Dnd 36 contract

Maxenboy:
Contract for Flags

### DIFF
--- a/src/Atoms/Flag/Flag.types.ts
+++ b/src/Atoms/Flag/Flag.types.ts
@@ -1,0 +1,9 @@
+import React from 'react';
+
+export type Props = {
+  height?: number;
+  width?: number;
+  country: string;
+};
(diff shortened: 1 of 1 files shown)


src/Atoms/Flag/Flag.types.ts
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export type Props = {
+  height?: number;
+  width?: number;
+  country: string;
```
mjmonline: flags should for a11y probably have a title prop

mjmonline: maybe the titles could be hardcoded in english to begin with

Maxenboy: will do that internally with the svg's

PR #55, opened 2019-04-08
Dnd 37 implementation

Maxenboy:
Flags :)


src/Atoms/Flag/Flag.tsx
```diff
@@ -0,0 +1,18 @@
...
+import flags from './flags';
+import { FlagComponent, Props } from './Flag.types';
+
+const StyledFlag = styled.svg<Props>`
+  width: ${p => p.theme.spacing.unit(p.width || 5)}px;
+  height: ${p => p.theme.spacing.unit(p.height || 5)}px;
+`;
+
+export const Flag: FlagComponent = props => {
+  const Component = flags[props.country.toLowerCase()];
```
shmidt-i: Any check here if the country is presented? For sure, typescript will catch it type-wise, but we still have a lot of consumers that don't use typescript, so runtime check should be there too?

Maxenboy: fixed

PR #58, opened 2019-04-08
Dnd 121 implementation

Maxenboy:
TitledCard :)


src/Molecules/TitledCard/TitledCard.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const TitledCard: TitledCardComponent = props =>
+  props.title ? (
+    <Card>
+      <StyledTitle>
+        <Text.Title3>{props.title}</Text.Title3>
+      </StyledTitle>
+      {props.children}
+    </Card>
+  ) : (
+    <Card>{props.children}</Card>
```
jauniuse: It feels for me that something is missing here.
As a consumer of this component I would want to use `TitledCard` in a same way as I would use `Card`, just with additional prop `title`, but this implementation restricts using `as` prop on `Card` for example. So such props need to be re-passed to `Card`.

We will have other components, like `CollabsibleCard` or `CardWithTabs`, but then they will need to have corresponding versions with titles also? Like `TitledCollabsibleCard`and `TitledCardWithTabs` - feels a bit too much. Maybe we should consider adding `title` prop to `Card` itself?

Maxenboy: fixed

src/Molecules/TitledCard/TitledCard.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import styled from 'styled-components';
+import R from 'ramda';
+import { Text, Card } from '../..';
+import { TitledCardComponent, Props } from './TitledCard.types';
+
+const StyledTitle = styled.div<Props>`
+  margin: ${({ theme }) => theme.spacing.unit(5)}px;
```
shmidt-i: I think StyledTitle should only do margin-bottom, other spacing should be part of card

shmidt-i: So maybe 
```
const StyledCard = styled(Card)`
  padding: ${({ theme }) => theme.spacing.unit(5)}px;
`

shmidt-i:
And also, can we rename the component to `CardWithTitle` please?

src/Molecules/TitledCard/TitledCard.tsx
Maxenboy: fixed

PR #78, opened 2019-04-15
add titles to list

jauniuse:
Added left and right titles to list, moved to molecules


src/Molecules/List/List.types.ts
```diff
@@ -1,4 +1,6 @@
 export type Props = {
   children: React.ReactNode;
   as?: 'ul' | 'ol';
+  leftTitle?: string;
```
shmidt-i: Should be node? Or how we can pass in typography?

shmidt-i: I mean separate typography for titles

src/Molecules/List/List.stories.tsx
```diff
@@ -0,0 +1,49 @@
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import ListItem from '../../Atoms/ListItem';
```
shmidt-i: I understand that you are only moving stuff, but please change imports of components to be from '../..'

src/Molecules/List/List.types.ts
jauniuse: change type to `React.ReactNode`

src/Molecules/List/List.stories.tsx
jauniuse: updated

src/Molecules/List/List.tsx
```diff
@@ -0,0 +1,43 @@
...
+export const List: React.FunctionComponent<Props> = ({
+  as = 'ul',
+  leftTitle,
+  rightTitle,
+  children,
+}) => {
+  const showTitle = leftTitle || rightTitle;
+
+  return (
+    <>
```
shmidt-i: Should it be a Fragment though? 🧐

jauniuse: it's now in `div` if title show be shown. I added `className` as well though that goes only to `ul`. If a container itself needs to be styled, it can be wrapped from outside with another element

jauniuse:
closing since it was decided to create a new component for list with titles

PR #79, opened 2019-04-15
add number, percent and time

jauniuse:
moved from mutations


src/Molecules/Percent/Percent.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { Number as NumberComponent } from '../..';
+import { PercentComponent as PercentComponentType } from './Percent.types';
+
+const Percent: PercentComponentType = ({ value, decimals = 0, invalidValue = '-' }) =>
+  value && Number.isFinite(value) ? (
+    <>
+      <NumberComponent value={value} decimals={decimals} />%
```
JacobBlomgren: `intl.formatNumber` has percentage options and I would rather use that and see that as our "reusable number component"

jauniuse: ehm.. That's right. But then need to do number formatting again here. Would either extract decimals formatting options to a separate part (I think it would just make things more complicated) or copy them here which would not be good. 

This approach I think is a bit more simple, we let `Number` component to handle number formatting and here we just add `%` sign.

JacobBlomgren: But we need decimals formatting also in `Currency` and `Development`. And those can't properly use `Number`

PR #81, opened 2019-04-16
fix: add checks for allowed weights and add missing typography types

JacobBlomgren:


src/Atoms/Typography/Typography.tsx
```diff
@@ -29,15 +30,19 @@ export const TYPOGRAPHY_TYPES: Record<Types, Types> = {
...
   title3: 'title3',
+  hero: 'hero',
 };
 
 const getTypeStyles = (props: ThemedStyledProps<Props, Theme>) => {
   const { type, weight, theme } = props;
   let sizeMobile;
   let sizeDesktop;
   let defaultWeight;
+  let allowedWeights = ['regular', 'bold', 'extrabold'];
```
mjmonline: can we do this with ts instead?

JacobBlomgren: Ivan and I tried really hard. Added as todo

PR #83, opened 2019-04-17
add list with titles

jauniuse:


src/Molecules/ListWithTitles/ListWithTitles.stories.tsx
```diff
@@ -0,0 +1,79 @@
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import { Flexbox, ListItem } from '../..';
+import ListWithTitles from '.';
```
shmidt-i: Import this from '../..' as well

src/Molecules/ListWithTitles/ListWithTitles.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const ItemContainer: React.FunctionComponent<{ children: React.ReactNode }> = ({ children }) => (
```
shmidt-i: You can skip children here, they are already part of React.FunctionComponent

src/Molecules/ListWithTitles/ListWithTitles.stories.tsx
jauniuse: updated

src/Molecules/ListWithTitles/ListWithTitles.stories.tsx
jauniuse: updated

PR #86, opened 2019-04-19
feat: replace with dumbed down spinner

mjmonline:
This PR changes the contract, old contract does not make sense with new implementation


shmidt-i:
That's just beautiful!!!!! 👏👏👏

PR #95, opened 2019-04-23
Dnd 152 remove padding titled card

Maxenboy:


shmidt-i:
Gonna merge tabs, which contains removal of padding in cardwithtitle

PR #104, opened 2019-04-25
Feat: link buttons

mjmonline:


src/Molecules/Button/Button.stories.tsx
```diff
@@ -119,4 +120,28 @@ storiesOf('Molecules | Button', module)
...
       </Flexbox.Item>
     </Flexbox.Container>
+  ))
+  .add('Link looking like a button', () => (
+    <Display
+      items={[
+        {
+          title: 'Reset',
+          component: (
+            <BrowserRouter>
```
JacobBlomgren: minor, minor, but how does this work with storybook? Maybe hashrouter?

src/Molecules/Button/Button.tsx
```diff
@@ -45,42 +48,79 @@ const getBackgroundColor = (props: ThemedStyledProps<Props, Theme>) => {
...
 `;
 
 export const Button: ButtonComponent = props => {
+  const typeIsNotPresent = typeof props.type === 'undefined';
+  const { disabled, onClick, size, type = 'button', variant, fullWidth, to, children } = props;
+  const toAndDisabledAreNotPresentTogether = !(to && disabled);
+
+  assert(
+    toAndDisabledAreNotPresentTogether,
+    "Button: You're using `to` prop together with `disabled` prop. Link's can't be disabled",
```
JacobBlomgren: Can't they? Are we sure instances of this won't come up?

PR #113, opened 2019-04-29
feat: new number and development components

JacobBlomgren:


src/Atoms/Number/Number.tsx
```diff
@@ -2,16 +2,36 @@ import React from 'react';
...
+}) => {
+  if (typeof value === 'undefined' || value === null || !Number.isFinite(value)) return <>-</>;
+  if (typeof currency !== 'undefined' && currency === null) return <>-</>;
+
+  return (
+    <>
+      {getPrefix(sign, value)}
+      {intl.formatNumber(value, getNumberOptions(decimals))}
+      {percentage && '%'}
+      {currency && ` ${currency}`}
```
mjmonline: On Account overview page we have currency in a smaller typography than the number. So they can be different font size. Can we instead let the developer pass in `< Typography type="primary">SEK</Typography>` into `currency` prop? Maybe same for the rest of the props?

src/Molecules/Development/Development.tsx
```diff
@@ -6,33 +6,24 @@ import { Theme } from '../../theme/theme.types';
 
 const getPrefix = (value?: number | null) => {
   if (!value) return '';
-  if (value > 0) return '▲';
-  return '▼';
+  return value > 0 ? '▲' : '▼';
```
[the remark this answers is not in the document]
JacobBlomgren: In the future yes, but not part of this refactor, I think.

src/Atoms/Number/Number.tsx
JacobBlomgren: Oh good point. Which other props do you mean though?

PR #120, opened 2019-05-02
globe icon, tabsnav fix

jauniuse:


src/Molecules/TabsNav/TabsNav.tsx
```diff
@@ -46,7 +46,7 @@ const StyledUl = styled.ul`
   margin-bottom: -1px;
 `;
 const isItemElement = (x: any): x is { type: typeof Item; props: ItemProps } =>
-  x != null && typeof x === 'object' && Object.hasOwnProperty.call(x, 'type') && x.type === Item;
+  x != null && typeof x === 'object' && Object.hasOwnProperty.call(x, 'type'); // FIXME: && x.type === Item;
```
jauniuse: this is the same issues as per https://github.com/nordnet/ui/blob/beta/src/Molecules/Tabs/Tabs.tsx#L63

PR #121, opened 2019-05-02
feat: update typography styles and add line height

mjmonline:


src/Atoms/Typography/Typography.tsx
```diff
@@ -41,57 +41,77 @@ const getTypeStyles = (props: ThemedStyledProps<Props, Theme>) => {
...
+      sizeMobile = 24;
+      sizeDesktop = 28;
+      lineHeightMobile = 7;
+      lineHeightDesktop = 8;
       defaultWeight = 'extrabold';
       break;
     case TYPOGRAPHY_TYPES.title2:
       sizeMobile = 20;
       sizeDesktop = 24;
+      lineHeightMobile = 6;
```
jesperwiner: These seems weird to me, some number thats multiplied and divided later on.
Size above is in pixels and this value is in number of spacing-units

mjmonline: I made a effort to clear up the calculation a little please check that. I am trying to give a unitless line-height to each typography style. To get this multiplier number we divide with the font-size. It's just a better unit to use for line-height

jesperwiner: Its more text but I like it better

PR #124, opened 2019-05-02
match route exact

jauniuse:


src/Molecules/TabsNav/TabsNav.tsx
```diff
@@ -58,7 +58,9 @@ const TabsNav: Component = (withRouter(({ children, location }) => {
     if (!isItemElement(c)) {
       assert(false, 'There should be only <TabsNav.Tab> children inside of <TabsNav> component');
     } else {
-      const isIndexActive = Boolean(matchPath(location.pathname, c.props.to));
+      const isIndexActive = Boolean(
+        matchPath(location.pathname, { path: c.props.to, exact: true }),
```
shmidt-i: Should it be able to have `exact` as a prop?

jauniuse: I can add it as a prop, though I think it should default to `true` then

jauniuse: updated

PR #127, opened 2019-05-03
fix: unique spinner ids

JacobBlomgren:
Also ran prettier, which was really needed (🤷‍♂️) but it should be in separate commits.

### DIFF
--- a/src/Atoms/Spinner/Spinner.tsx
+++ b/src/Atoms/Spinner/Spinner.tsx
@@ -19,12 +19,11 @@ const StyledSvg = styled.svg`
   display: block;
 `;
 
-const RawSpinner: React.FC<PropsWithTheme> = ({ theme, size, color }) => {
+const RawSpinner: React.FC<PropsWithTheme> = ({ theme, size, color, id }) => {
   const calculatedSize = theme.spacing.unit(size || 4);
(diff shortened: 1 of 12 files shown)


src/Atoms/Spinner/Spinner.tsx
```diff
@@ -19,12 +19,11 @@ const StyledSvg = styled.svg`
...
 `;
 
-const RawSpinner: React.FC<PropsWithTheme> = ({ theme, size, color }) => {
+const RawSpinner: React.FC<PropsWithTheme> = ({ theme, size, color, id }) => {
   const calculatedSize = theme.spacing.unit(size || 4);
   const usedColor = typeof color === 'function' ? color(theme) : color || theme.color.cta;
-  const sanatizedColor = usedColor.replace(/#\s?/g, '');
-  const id1 = `spinner-${sanatizedColor}-1`;
-  const id2 = `spinner-${sanatizedColor}-2`;
+  const id1 = `spinner-${id}-1`;
```
jauniuse: is there no other way to generate id implicitly in spinner? :>

JacobBlomgren: no we decided on this for now, it's really hard 😢

jauniuse: :( it so complicates usage of `Spinner`

PR #154, opened 2019-05-10
fix: media max-width issue

mjmonline:


mjmonline:
`max-width` in media query should be 1 pixel less then `min-with` or else both will be true at specified breakpoint

PR #162, opened 2019-05-13
Feat/flexbox redone

shmidt-i:


src/Atoms/Flexbox/Flexbox.stories.tsx
```diff
@@ -13,69 +13,107 @@ const StyledContent = styled.div`
...
+      </Flexbox>
+      <Flexbox item size={6}>
+        <Content>Col 2</Content>
+      </Flexbox>
+      <Flexbox item size={3}>
+        <Content>Col 3</Content>
+      </Flexbox>
+    </Flexbox>
   ))
   .add('Flexbox with custom gutter', () => (
```
mjmonline: maybe not custom gutter because it seems there is no default gutter

src/Atoms/Flexbox/Flexbox.stories.tsx
```diff
@@ -13,69 +13,107 @@ const StyledContent = styled.div`
...
-        <Flexbox.Item order={1}>
+        <Flexbox item order={1}>
           <Content>Col 2</Content>
-        </Flexbox.Item>
-      </Flexbox.Container>
+        </Flexbox>
+      </Flexbox>
+    );
+  })
+  .add('Flexbox container item', () => {
```
mjmonline: 🎉

mjmonline: maybe `with combination of container and item`

mjmonline:
Add support to use className

PR #167, opened 2019-05-15
Add select

JacobBlomgren:


src/Molecules/Select/Select.tsx
```diff
@@ -0,0 +1,133 @@
...
+  className,
+  value: valueFromProps,
+  onChange: onChangeFromProps,
+  onBlur: onBlurFromProps,
+  onFocus: onFocusFromProps,
+}) => {
+  const [focus, setFocus] = useState(false);
+  const [value, setValue] = useState();
+
+  const onChange = e => {
```
JacobBlomgren: just need typescript help here @shmidt-i

mjmonline:
Great!  🙌

mjmonline:
We might need to have a prop to make the select higher/bigger

src/Molecules/Select/Select.tsx
jauniuse: I suppose e is of type `React.ChangeEvent<HTMLSelectElement>`

PR #179, opened 2019-05-16
Max and min decimals

JacobBlomgren:


src/Molecules/Development/__snapshots__/Development.stories.storyshot
```diff
@@ -1454,6 +1454,50 @@ exports[`Storyshots Molecules | Development Regression: value is non-zero, but r
...
+  >
+    <div
+      className="c2"
+    >
+      \`value = -0.3\`
+    </div>
+    <div>
+      <span
+        className="c3"
+        value={-0}
```
shmidt-i: `value` got propagated to span's attributes

JacobBlomgren: I think it has for a while though, and we've had this problem for quite some time with other components, but yes you're right 😞

PR #183, opened 2019-05-17
Select fix

JacobBlomgren:
* Chevron was not clickable
* There was an unnecessary span for screen readers announcing what the transparent select already did.

### DIFF
--- a/src/Atoms/Typography/Typography.tsx
+++ b/src/Atoms/Typography/Typography.tsx
@@ -1,9 +1,8 @@
 import React from 'react';
 import styled, { ThemedStyledProps } from 'styled-components';
-import * as R from 'ramda';
 import { Theme } from '../../theme/theme.types';
 import { Props, Types } from './Typography.types';
-import { assert } from '../../common/utils';
+import { assert, pickAriaAttributes } from '../../common/utils';
 
 const WEIGHTS = {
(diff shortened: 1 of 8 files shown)


src/Atoms/Typography/Typography.tsx
```diff
@@ -146,8 +147,17 @@ const StyledTypography = styled.span<Props>`
 export const Typography: React.FC<Props> = props => {
   const { as, className, children, color, type, weight } = props;
 
+  const arias = R.pickBy((_, key: string) => R.test(/^aria-/, key), props);
```
shmidt-i: Can we move it in /utils and also, maybe test coverage?

src/Molecules/Select/__snapshots__/Select.stories.storyshot
```diff
@@ -1105,6 +1104,7 @@ exports[`Storyshots Molecules | Select Select without placeholder 1`] = `
       </option>
     </select>
     <span
+      aria-hidden={true}
```
shmidt-i: is this a correct value for html? shouldn't it be string `"true"`? not sure

JacobBlomgren: it becomes that in the end

PR #188, opened 2019-05-20
Fix -0 attempt 2

JacobBlomgren:
So it turned out that react intl does something weird to the '-' sign

### DIFF
--- a/src/Molecules/Number/Number.tsx
+++ b/src/Molecules/Number/Number.tsx
@@ -108,10 +108,11 @@ const NumberComponent: NumberComponentType = ({
     getNumberOptions(value, { ticks, decimals, minimumDecimals, maximumDecimals }),
   );
 
+  const isMinusZero = (val: number) => 1 / val === -Infinity;
   const number = (
     <>
       {getPrefix(sign, roundedValue)}
-      {formattedNumber !== '-0' ? formattedNumber : '0'}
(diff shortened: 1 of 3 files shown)


Gysis:
the -0 check seems a bit weird. Isn't there a sign() function you can use?

JacobBlomgren:
> isMinusZero = val === 0 && Math.sign(val) === -1
> 
> i thoink would be more clear

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/sign
> Math.sign(-0);    // -0
I'm afraid not

src/Molecules/Number/Number.tsx
```diff
@@ -108,10 +108,11 @@ const NumberComponent: NumberComponentType = ({
     getNumberOptions(value, { ticks, decimals, minimumDecimals, maximumDecimals }),
   );
 
+  const isMinusZero = (val: number) => 1 / val === -Infinity;
```
shmidt-i: Is it legal is Sweden ? 😂

JacobBlomgren: what do you mean?

shmidt-i: dividing by zero :) anyway, it feels like a decent check here :)

PR #190, opened 2019-05-21
fix: correct select height

JacobBlomgren:
just fixing a slip up


src/Molecules/Select/Select.tsx
```diff
@@ -4,12 +4,12 @@ import styled from 'styled-components';
 import { VisuallyHidden, Icon, Typography } from '../..';
 import { SelectComponent } from './Select.types';
 
-const SelectHeight = 8;
-const ArrowSpace = 7;
+const selectHeight = 8;
```
shmidt-i: Should be CAPITALIZED_SNAKE_CASE, isn't it?

JacobBlomgren: Yup

PR #199, opened 2019-05-27
Add callback to collapsible card onClick

Maxenboy:


src/Molecules/CollapsibleCard/CollapsibleCard.tsx
```diff
@@ -145,12 +147,22 @@ export const CollapsibleCard: React.FC<CollapsibleProps> = ({
...
+    onTouchStart(e);
+  };
+
   return (
     <Card>
       <StyledButton
         type="button"
-        onClick={toggle}
-        onTouchStart={toggle}
+        onClick={onClickCallback}
```
JacobBlomgren: since `toggle` is called from both `onClick` and `onTouchStart`, we could just call the `onClick` from props callback there and not expose `onTouchStart`. That's like an internal detail.

PR #206, opened 2019-05-28
fix(Flag): check for undefined country

JacobBlomgren:


mjmonline:
@JacobBlomgren I guess Vlad talked to both of us :) I have a PR doing almost the same thing

JacobBlomgren:
> @JacobBlomgren I guess Vlad talked to both of us :) I have a PR doing almost the same thing

I like yours better :)

PR #214, opened 2019-05-31
feat: add flexbox screen sized props

shmidt-i:


src/Atoms/Flexbox/Flexbox.tsx
```diff
@@ -40,9 +40,11 @@ const getGutterStyles = ({
 }) => {
   return direction === 'row' || direction === 'row-reverse'
     ? `
-  margin-left: ${theme.spacing.unit(gutter)}px;
-`
+    margin-top: 0;
```
mjmonline: do we need this?

src/Atoms/Flexbox/Flexbox.tsx
```diff
@@ -40,9 +40,11 @@ const getGutterStyles = ({
...
+    margin-left: ${theme.spacing.unit(gutter)}px;
+    `
     : `
+      margin-left: 0;
```
mjmonline: do we need this?

src/Atoms/Flexbox/Flexbox.tsx
mjmonline: nvm :)

src/Atoms/Flexbox/Flexbox.tsx
mjmonline: nvm :)

shmidt-i: That what I am afraid of. If for some reason you want to have that margin-left, would you be able to by doing styled(Flexbox) ?

PR #218, opened 2019-06-03
Disabled button fix

Maxenboy:
@shmidt-i @mjmonline pls take a look and feel free to fix if I've forgotten anything

### DIFF
--- a/src/Molecules/Button/Button.stories.tsx
+++ b/src/Molecules/Button/Button.stories.tsx
@@ -228,6 +228,20 @@ storiesOf('Molecules | Button', module)
       </Button>
     </HashRouter>
   ))
+  .add('Disabled button', () => (
+    <HashRouter>
+      <Button
+        fullWidth
+        onClick={action('clicked')}
+        size="m"
+        to="www.google.com"
+        variant="secondary"
+        disabled
+      >
+        Button
+      </Button>
+    </HashRouter>
+  ))
   .add('Link looking like a button with with colors', () => (
     <HashRouter>
       <>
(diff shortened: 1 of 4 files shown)


src/Molecules/Button/Button.stories.tsx
```diff
@@ -228,6 +228,20 @@ storiesOf('Molecules | Button', module)
       </Button>
     </HashRouter>
   ))
+  .add('Disabled button', () => (
```
mjmonline: On line 18 we have this title already, maybe use `Disabled link looking like a button`

src/Molecules/Button/Button.tsx
```diff
@@ -126,7 +126,7 @@ const getSharedStyle = (props: ThemedStyledProps<ButtonProps | LinkProps, Theme>
   return `
     ${getBackgroundColor(props)}
     box-sizing: border-box;
-    border: ${BORDER_SIZE}px solid ${getColorWithDefault('transparent')};
+    border: ${disabled ? 'none' : `${BORDER_SIZE}px solid ${getColorWithDefault('transparent')}`};
```
mjmonline: We still want a transparent border here to maintain the height between different looks

src/Molecules/Button/Button.tsx
```diff
@@ -162,13 +162,6 @@ export const Button: ButtonComponent = props => {
     rel,
     color,
   } = props;
-  const toAndDisabledAreNotPresentTogether = !(to && disabled);
-
-  assert(
-    toAndDisabledAreNotPresentTogether,
-    "Button: You're using `to` prop together with `disabled` prop. Link's can't be disabled",
```
mjmonline: @shmidt-i what do you think of removing this warning?

shmidt-i: Maybe not removing, but instead changing the text to
```
Button: You're using `to` prop together with `disabled` prop. `Disabled` prop won't be propagated to the dom node, because <a> element can't be disabled 
```

PR #219, opened 2019-06-04
feat(spinner): add delay to show spinner

jesperwiner:
A suggestion, we have this on graph component that if the data is loaded fast enough that the user is not even thinking about pressing F5 again yet we don't show the spinner at all.

### DIFF
--- a/src/Atoms/Spinner/Spinner.types.ts
+++ b/src/Atoms/Spinner/Spinner.types.ts
@@ -8,7 +8,7 @@ export type Props = {
   color?: ColorFn;
   /** Globally unique id for the spinner */
   id: string;
-  instant?: boolean;
+  delay?: boolean | number;
 };
 
 export type PropsWithTheme = Props & {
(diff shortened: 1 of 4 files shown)


src/Atoms/Spinner/Spinner.types.ts
```diff
@@ -8,6 +8,7 @@ export type Props = {
   color?: ColorFn;
   /** Globally unique id for the spinner */
   id: string;
+  instant?: boolean;
```
shmidt-i: I see the point and I like the idea, but let's maybe do it in a non-breaking way?

How about to have
```
delay?: boolean | number
```
And consume it like that:
```
<Spinner delay /> // would result in default delay
<Spinner delay={1000} /> //would result in 1s delay
```
What do you think @jesperwiner @iamstarkov ?

shmidt-i: And ofc we will need to manually propagate changes to pages then, but imo it's better this way

shmidt-i: Also, `instant` is kinda confusing name here

PR #220, opened 2019-06-04
feat: create switch component

mjmonline:


src/Molecules/Switch/Switch.tsx
```diff
@@ -0,0 +1,126 @@
...
+      border: 1px solid ${p => p.theme.color.disabledBackground};
+    }
+  }
+`;
+
+const generateSomewhatUniqueId = (str: string) => {
+  const safeString = str.replace(/^[^a-z]+|[^\w:.-]+/gi, '').toLowerCase();
+
+  return `switch-${safeString}`;
+};
```
iamstarkov: maybe use https://github.com/ai/nanoid instead

src/Molecules/Switch/Switch.types.ts
```diff
@@ -0,0 +1,7 @@
+export type Props = {
+  className?: string;
+  defaultOn?: boolean;
```
shmidt-i: What about controllable vs uncontrollable? 
Currently this component is 100% uncontrollable, which may hit us in some time, we probably want both behaviours to exist

src/Molecules/Switch/Switch.tsx
mjmonline: I checked up the use of <label> outside <form> and it seems ok. So now I wrap the switch in a <label> which allows me to not need any id

src/Molecules/Switch/Switch.types.ts
mjmonline: @shmidt-i added support for this

src/Molecules/Switch/Switch.tsx
iamstarkov: that's good news

src/Molecules/Switch/Switch.types.ts
```diff
@@ -0,0 +1,12 @@
+export type Props = {
+  className?: string;
+  checkedInitially?: boolean;
+  /**
+   * Using this prop will enable controlled behaviour
+   * https://reactjs.org/docs/forms.html#controlled-components
+   */
+  checked?: boolean;
+  disabled?: boolean;
+  labelText: string;
```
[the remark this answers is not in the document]
mjmonline: Done

PR #221, opened 2019-06-07
feat: improved media and useMedia hook

shmidt-i:


src/Atoms/Media/Media.tsx
```diff
@@ -15,4 +16,33 @@ const StyledDiv = styled.div<{query: Props['query']}>`
...
-export const Media: React.FunctionComponent<Props> = props => <StyledDiv {...props} />;
+const noop = () => {};
+const useMedia = (query: string | ((t: Theme) => string)) => {
+  const theme = React.useContext(ThemeContext);
+  const mediaQuery = (typeof query === 'string' ? query : query(theme)).replace('@media ', '');
+
+  const mediaHit =
+    typeof window === 'undefined' || !window || !window.matchMedia
+      ? { matches: null, addListener: noop, removeListener: noop }
+      : window.matchMedia(mediaQuery);
```
jauniuse: I think this one will be re-created each time hook is run, so effect below will be run on each render

image

Or im missing something here?

src/Atoms/Media/Media.tsx
```diff
@@ -15,4 +16,33 @@ const StyledDiv = styled.div<{query: Props['query']}>`
         display: none;
     }
 `
-export const Media: React.FunctionComponent<Props> = props => <StyledDiv {...props} />;
+const noop = () => {};
+const useMedia = (query: string | ((t: Theme) => string)) => {
+  const theme = React.useContext(ThemeContext);
+  const mediaQuery = (typeof query === 'string' ? query : query(theme)).replace('@media ', '');
+
+  const mediaHit =
```
jauniuse: I would write this code as:

```js
  const mediaHit = React.useRef({ matches: null, addListener: noop, removeListener: noop });
  const mediaQuery = (typeof query === 'string' ? query : query(theme)).replace('@media ', '');

  React.useEffect(() => {
      mediaHit = window.matchMedia(mediaQuery)
  }, [mediaQuery])

   const [matches, setMatches] = React.useState(mediaHit.matches);

  // ....
```

so `useEffect` is not being run on server, it will be `null` as per initial value in ref. `mediaHit` will
be preserved between renders as a same reference because of `useRef` unless `mediaQuery` changes.
Also with this approach usage of side effect (`window`) will be in `effect` instead of being directly in render

shmidt-i: Awesome, tack!

src/Atoms/Media/Media.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+    </ThemeProvider>,
+  );
+  expect(html).toMatchSnapshot();
+});
+
+test('Client-side rendering works: not rendering non-matched media', async () => {
+  const { getByTestId } = render(
+    <ThemeProvider theme={createTheme()}>
+      <>
+        <Media query={t => t.media.greaterThan(t.size.sm)}>
```
jauniuse: I think it would be nice to have a test case for when query prop changes, I mean something like:

```js
  const firstQuery = t => t.media.greaterThan(t.size.sm);
  const { rerender, getByTestId } = render(<Media query={firstQuery}>some content</Media>)
  
  // expect that content shows
  // expect(getByTestId('XXX')) ..

  const secondQuery = t => t.media.lessThan(t.size.sm);
  
  // rerender with different prop
  const { getByTestId } = rerender(<Media query={secondQuery}>some other content</Media> )

  // expect that content does not show
  
```

jauniuse: and then also can test if event listener was removed after prop change

jauniuse:
this is awesome 💓

src/Atoms/Media/Media.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const main = document.createElement('main');
+
+  main.innerHTML = html;
+  const htmlSSR = main.innerHTML;
+
+  ReactDOM.hydrate(<Element />, main);
+
+  const htmlBeforeEffect = main.innerHTML;
+
+  await new Promise(res => setTimeout(res, 0)); // wait for effect to be invoked
```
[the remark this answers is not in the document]
shmidt-i: not gonna do that, want to emphasis on that first render on client is gonna be the same as in ssr

src/Atoms/Media/Media.test.tsx
shmidt-i: Test listener - gonna skip for now, too much mocking, rerender test added

PR #225, opened 2019-06-12
Feat: add disabled state for links and also render buttons that looks like links

mjmonline:


src/Molecules/LinkSell/LinkSell.types.ts
```diff
@@ -1,9 +1,8 @@
...
-import { Props as LinkProps } from '../Link/Link.types';
+import { LinkProps, ButtonProps } from '../Link/Link.types';
 
 export type Props = {
-  disabled?: boolean;
   as?: string;
 };
 
-export type LinkSellComponent = React.FunctionComponent<LinkProps & Props>;
+export type LinkSellComponent = React.FunctionComponent<LinkProps | ButtonProps & Props>;
```
shmidt-i: `LinkProps | ButtonProps & Props` would be equal to `LinkProps | (ButtonProps & Props)`, probably not what you wanted :)

src/Molecules/LinkBuy/LinkBuy.types.ts
```diff
@@ -1,9 +1,8 @@
...
-import { Props as LinkProps } from '../Link/Link.types';
+import { LinkProps, ButtonProps } from '../Link/Link.types';
 
 export type Props = {
-  disabled?: boolean;
   as?: string;
 };
 
-export type LinkBuyComponent = React.FunctionComponent<LinkProps & Props>;
+export type LinkBuyComponent = React.FunctionComponent<LinkProps | ButtonProps & Props>;
```
shmidt-i: `LinkProps | ButtonProps & Props` would be equal to `LinkProps | (ButtonProps & Props)`, probably not what you wanted :)

src/Molecules/Link/__snapshots__/Link.stories.storyshot
```diff
@@ -90,14 +92,16 @@ exports[`Storyshots Molecules | Link Link with _blank target 1`] = `
 exports[`Storyshots Molecules | Link Link with rel as nofollow 1`] = `
 .c0 {
   font-family: 'Nordnet Sans Mono',-apple-system,BlinkMacSystemFont,'Segoe UI',Roboto,sans-serif;
-  color: #0046FF;
+  color: #282823;
```
shmidt-i: Why this color changed?

mjmonline: Because with this PR we are doing `color: ${disabled ? theme.color.disabledText : theme.color.cta};` inside the component. With that approach setting the color from the outside won't override the rule inside the compoent. So then that color is redundant to set in all examples, so I removed it

src/Molecules/LinkSell/LinkSell.types.ts
mjmonline: fixed

src/Molecules/LinkBuy/LinkBuy.types.ts
mjmonline: fixed

PR #226, opened 2019-06-13
[WIP] Feat: Tabs and TabsNav revisit

mjmonline:


src/Molecules/Tabs/Tabs.stories.tsx
```diff
@@ -51,6 +51,18 @@ storiesOf('Molecules | Tabs', module)
       </StyledTabs>
     </Typography>
   ))
+  .add('With height modified', () => (
+    <Typography type="secondary">
+      <Tabs height={11}>
```
shmidt-i: I think this is too much freedom to screw things up :)
Maybe better to fixate the amount of possible sizes?

mjmonline: @shmidt-i the intension here was to have a solution for when we need tabs to fit a certain container height.

PR #227, opened 2019-06-13
fix: flexbox component cleanup

mjmonline:


src/Atoms/Flexbox/Flexbox.stories.tsx
```diff
@@ -117,9 +117,16 @@ storiesOf('Atoms | Flexbox', module)
       </Flexbox>
     );
   })
-  .add('Flexbox depending on screen size', () => {
+  .add('With media query props', () => {
```
shmidt-i: Media query = implementation details? 
Screen size or breakpoints is probably better fit here?

PR #231, opened 2019-06-17
Feat: update grid and breakpoints

mjmonline:


src/theme/Theme.stories.tsx
```diff
@@ -71,9 +72,9 @@ storiesOf('Theme', module)
   })
   .add('Screen sizes', () => (
     <Display
-      items={Object.entries(theme.size).map(([title, size]) => ({
+      items={Object.entries(theme.breakpoints).map(([title, breakpoint]) => ({
```
shmidt-i: here I think we can display both size and offset

src/theme/theme.ts
```diff
@@ -45,17 +46,50 @@ export const rawColor = {
...
+    size: 1440,
+    offset: 24,
+  },
+  xl: {
+    size: 1680,
+    offset: 10,
+  },
+};
+
+const size: Theme['size'] = {
```
shmidt-i: Let's write `deprecated` helper and use it here

src/theme/theme.types.ts
```diff
@@ -130,22 +145,20 @@ export type Theme = {
...
-`
+     * styled.div`${({ theme }) => theme.media.lessThan(theme.breakpoints.md)} {}`
      */
-    lessThan: (/** One of theme.size */ size: number) => MediaQuery;
-    greaterThan: (/** One of theme.size */ size: number) => MediaQuery;
-    between: (
-      /** One of theme.size */ size1: number,
-      /** One of theme.size */ size2: number,
-    ) => MediaQuery;
+    lessThan: (number: NumberOrObjectWithNumber) => MediaQuery;
```
shmidt-i:
```suggestion
    lessThan: (size: NumberOrObjectWithNumber) => MediaQuery;
```

src/theme/theme.types.ts
```diff
@@ -130,22 +145,20 @@ export type Theme = {
...
+    greaterThan: (number: NumberOrObjectWithNumber) => MediaQuery;
```
shmidt-i:
```suggestion
    /**
     * @param size One of theme.breakpoints
     * @example
     * styled.div`${({ theme }) => theme.media.greaterThan(theme.breakpoints.md)} {}`
     */
    greaterThan: (size: NumberOrObjectWithNumber) => MediaQuery;
```

src/theme/theme.ts
shmidt-i:
```js
const deprecate = message => target => {
  if (process.env.NODE_ENV !== "production") {
    return typeof Proxy === "undefined"
      ? target
      : new Proxy(target, {
          get: function(getTarget, getProp) {
            console.warn(`Deprecated: ${message}`);
            return getTarget[getProp];
          },
          apply: function(applyTarget, thisArg, argumentsList) {
            console.warn(`Deprecated: ${message}`);
            return applyTarget.apply(thisArg, argumentsList);
          }
        });
  }
  return target;
};

```

shmidt-i: Something like that

src/theme/Theme.stories.tsx
mjmonline: fixed

src/theme/theme.ts
mjmonline: fixed

src/theme/theme.types.ts
mjmonline: fixed

src/theme/theme.types.ts
mjmonline: fixed

PR #240, opened 2019-06-26
Feat: update grid breakpoints

mjmonline:


src/common/utils.ts
```diff
@@ -20,7 +20,26 @@ export const assert = (
   return true;
 };
 
+export const deprecate = (message: string) => <T extends {} | Function>(target: T): T => {
+  if (process.env.NODE_ENV !== 'production') {
+    return typeof Proxy === 'undefined'
```
shmidt-i: Let's check the cage against IE11 and Firefox just to be sure

src/theme/theme.ts
```diff
@@ -45,17 +46,47 @@ export const rawColor = {
...
   const { a11yColors = false } = config;
-  const size: Theme['size'] = {
-    xs: 360,
-    sm: 768,
-    md: 992,
-    lg: 1300,
-  };
 
-  const sizeValues = Object.values(size) as number[];
+  // @ts-ignore
```
shmidt-i: is it still needed if you are casting to number[]?

mjmonline: Expected 0 arguments, but got 1

PR #241, opened 2019-06-27
Feat/dnd 172 cssgrid conditional hiding

shmidt-i:


src/Atoms/CssGrid/CssGrid.tsx
```diff
@@ -234,19 +272,20 @@ const CssGridItem: React.FC<ItemProps> = ({ align, area, children, justify, plac
...
-  const info = areasInfo[areaName];
-  const styles = [`-ms-grid-row: ${info.rowStart};`, `-ms-grid-column: ${info.colStart};`];
-  styles.push(`-ms-grid-column-span: ${info.colSpan};`);
-  styles.push(`-ms-grid-row-span: ${info.rowSpan};`);
+const generateChildStyles = (areasInfo: Record<string, AreaInfo>, size: Size, theme: Theme) => (
+  areaName: string,
+): string => {
+  const styles = [];
+  if (!areasInfo[areaName]) {
+    styles.push('display: none; grid-area: none;');
```
mjmonline: I'm not sure we should diverge from the real css grid implementation. If a area is missing it should simply get thrown outside the grid and the developer can then manually hide it for that specific breakpoint. If we do manage to get the css grid code autoprefixed one day then our implementation would differ and thus making it harder to migrate

PR #244, opened 2019-07-02
Fix: flexbox gutters

mjmonline:


src/Atoms/Flexbox/Flexbox.tsx
```diff
@@ -60,7 +67,14 @@ const getContainerStyles = (p: Props & { theme: Theme }) => `
   ${
     p.gutter
       ? `
-    & > *:not(:first-child) {
+    ${
+      p.direction === 'column' || p.direction === 'column-reverse'
```
Maxenboy: why is this only applied for direction column and column-reverse?
The situation we are in, the items are on a row with wrap so the direction row also needs negative margins I guess

shmidt-i: Yes, let's add then all margins for all directions

shmidt-i:
What's with this one @mjmonline ?

mjmonline:
@shmidt-i having margins all around will cause collapsing margins. Which means that when the developer uses `<Box />` or something else with margins around the flexbox it would be 10px less than what they specify. Maybe that's ok if everyone knows about this, maybe we could have this as a prop as well. What do you think?

src/Atoms/Flexbox/Flexbox.tsx
```diff
@@ -64,22 +47,7 @@ const getContainerStyles = (p: Props & { theme: Theme }) => `
...
-        : `margin: 0 -${p.theme.spacing.unit(p.gutter / 2)}px;`
-    }
-
-    & > * {
-      ${getGutterStyles(p)}
-    }
-  `
-      : ''
-  }
+  ${p.gutter ? getGutterStyles(p) : ''}
```
shmidt-i: if you are checking here that `p.gutter` exists, then you don't need optional on line 32 and you don't need to use `!` at line 34 and 37

PR #245, opened 2019-07-03
fix: stop using theme.size internally

mjmonline:
This to stop all the warnings about it

### DIFF
--- a/src/Atoms/Media/Media.test.tsx
+++ b/src/Atoms/Media/Media.test.tsx
@@ -16,7 +16,7 @@ beforeEach(() => {
     dispatchEvent: () => null,
     onchange: () => null,
     removeListener: () => null,
-    matches: media === theme.media.greaterThan(theme.size.sm).replace('@media ', ''),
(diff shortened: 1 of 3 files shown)


mjmonline:
This to stop all the warnings about it

src/Atoms/Media/Media.test.tsx
```diff
@@ -16,7 +16,7 @@ beforeEach(() => {
     dispatchEvent: () => null,
     onchange: () => null,
     removeListener: () => null,
-    matches: media === theme.media.greaterThan(theme.size.sm).replace('@media ', ''),
+    matches: media === theme.media.greaterThan(theme.breakpoints.sm.size).replace('@media ', ''),
```
shmidt-i: But I recall we made this work with just `theme.media.greaterThan(theme.breakpoints.sm)`, am I right?

mjmonline: fixed

PR #250, opened 2019-07-10
fix: line below tabItem trimmed if overflow:hidden

shmidt-i:


src/Molecules/Tabs/Tabs.tsx
```diff
@@ -56,7 +56,7 @@ Title.displayName = 'Tabs.Title';
 
 const StyledUl = styled(List)`
   /** @todo check this out */
-  margin-bottom: -1px;
+  margin-bottom: 1px;
```
mjmonline: We should probably solve this with `::before` or `::after` to create the line and then remove this margin altogether

shmidt-i: It's already ::after with position: absolute. The thing is on page market overview they did a fix to prevent y scrolling which happens for some reason in mobile and ie11. We probably can solve this but not atm, let's merge this and iterate

PR #254, opened 2019-07-16
Feat: more icons and info molecules

KitsuneKelso:
Adds the `<FeedbackBanner />` component.
_"Feedback banner should be used within modules to display information that needs attention. Content is flexible, it can contain links and plain text."_

Comes in four variants:
Screenshot 2019-07-17 at 13 58 32


src/Molecules/InfoBox/InfoBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+    case 'info':
+    default:
+      return <Icon.InfoCircle size={5} fill={t => t.color.cta} />;
+  }
+};
+
+const StyledContainer = styled.div<InfoBoxProps>`
+  width: 100%;
+  min-height: 64px;
+  background: $colorWhite;
```
KitsuneKelso: Leftover from the migration, fix incoming!

mjmonline:
@KitsuneKelso Where is this component intended to be used? Because this looks similar to something designers called Feedback Banner. We also have something else called Info Bar which looks different than this.

KitsuneKelso:
@mjmonline yes, this is essentially the FeedbackBanner, I can rename the component to that if you would prefer.

src/Atoms/Icon/components/InfoCircle.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { IconBase } from '../IconBase';
+import { BaseProps } from '../IconBase.types';
+
+export const InfoCircle = (props: BaseProps) => {
+  return (
+    <IconBase {...props}>
+      <path
+        d="M12,24 C5.372583,24 0,18.627417 0,12 C0,5.372583 5.372583,0 12,0 C18.627417,0 24,5.372583 24,12 C24,18.627417 18.627417,24 12,24 Z"
+        id="ico-info"
```
mjmonline: Need this id?

src/Molecules/InfoBox/InfoBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+      return <Icon.CheckMarkCircle size={5} fill={t => t.color.positive} />;
+    case 'info':
+    default:
+      return <Icon.InfoCircle size={5} fill={t => t.color.cta} />;
+  }
+};
+
+const StyledContainer = styled.div<InfoBoxProps>`
+  width: 100%;
+  min-height: 64px;
```
mjmonline: please use `theme.spacing.unit(16)px`

src/Molecules/InfoBox/InfoBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+  background: ${p => p.theme.color.card};
```
mjmonline: FeedbackBanner has a gray background color from what I can see here https://nordnet.invisionapp.com/d/main#/console/16903964/365056550/preview

src/Molecules/InfoBox/InfoBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+  border-left: 4px solid ${getColor};
+  box-sizing: border-box;
+  display: flex;
```
mjmonline: You can use the `<Flexbox container />` component

src/Molecules/InfoBox/InfoBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+  align-items: center;
+`;
+
+const StyledIcon = styled.div`
+  margin-left: 16px;
```
mjmonline: If you use the Flexbox component on the container then the gutter prop would handle this

src/Molecules/InfoBox/InfoBox.tsx
```diff
@@ -0,0 +1,81 @@
...
+`;
+
+const StyledContent = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: column;
```
mjmonline: same here `<Flexbox container item />`

src/Molecules/InfoBox/InfoBox.types.ts
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export type InfoBoxProps = {
+  /** @default info */
+  variant?: 'info' | 'error' | 'warning' | 'success';
+  title?: string;
+  children: React.ReactChild | React.ReactChild[];
```
mjmonline: You don't have to define children if you have `React.FC< InfoBoxComponent >` on line 66 in InfoBox.tsx

src/Atoms/Icon/components/InfoCircle.tsx
KitsuneKelso: Missed this one when converting it from the SVG, it should be gone now!

src/Molecules/InfoBox/InfoBox.types.ts
KitsuneKelso: Didn't know that, thanks for the tip! :)

src/Molecules/InfoBox/InfoBox.tsx
KitsuneKelso: Wasn't able to resolve this one, when I applied the `gutter` the items stared bleeding outside the edges of the container horizontally. 🤔

KitsuneKelso: @mjmonline I realise that the problem is the negative margin that the gutter applies to the container: https://github.com/nordnet/ui/blob/master/src/Atoms/Flexbox/Flexbox.tsx#L34

Since the icon is the left-most child of the container, the padding it receives on the left side is negated by the negative margin on the parent. However, I'm guessing that this is the intended behaviour, it's only supposed to add a gutter between the items as the name implies, so I don't think it's going to work for this use-case.

mjmonline: @KitsuneKelso I pushed a commit, check it out, it was easier than explaining :)

src/Molecules/FeedbackBanner/FeedbackBanner.tsx
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const StyledContainer = styled.div<FeedbackBannerProps>`
+  background: ${getBackgroundColor};
+  border-left: ${p => p.theme.spacing.unit(1)}px solid ${getBorderColor};
+  padding: ${p => `${p.theme.spacing.unit(1)}px ${p.theme.spacing.unit(3)}px`};
+  box-sizing: border-box;
+`;
+
+const StyledContent = styled.div`
```
mjmonline: Probably you dont need this whole div as well. The overriding of the `a` tag is probably not necessary. We can just use `<Link />` and it would get the same style

src/Molecules/InfoBox/InfoBox.tsx
KitsuneKelso: Alright 👍

src/Molecules/FeedbackBanner/FeedbackBanner.types.ts
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+export type FeedbackBannerProps = {
+  /** @default info */
+  variant?: 'info' | 'error' | 'warning' | 'success';
+  /** @default gray */
+  backgroundColor?: 'gray' | 'white';
```
mjmonline: I'm not sure about having `backgroundColor` as strings. In the `<PageWrapper />` component we do something similar to this, there we provide a function check here:  https://github.com/nordnet/ui/blob/6b4a661a53e9f0fddaf536fe896065fa0bffa00c/src/Molecules/PageWrapper/PageWrapper.types.ts#L8

src/Molecules/FeedbackBanner/FeedbackBanner.tsx
KitsuneKelso: Good point, removing it!

src/Molecules/FeedbackBanner/FeedbackBanner.types.ts
KitsuneKelso: Also a good point, I'll adapt it so that it can grab any color from the theme. 🙂

KitsuneKelso: @mjmonline I saw the comment about moving the `ColorFn` out to util types, so I kept it the same and copied it over for now until there's time to move it out.

KitsuneKelso:
@mjmonline I have no idea what's going on with that Travis build, when I run the tests locally everything is fine, and I can't seem to re-trigger the tests here for some reason.
Screenshot 2019-07-17 at 13 50 58

KitsuneKelso:
@mjmonline Ok, now it makes sense, the changes in master were messing with the results in the snapshot. I rebased it and updated the storyshot. 🎉

mjmonline:
> @mjmonline I have no idea what's going on with that Travis build, when I run the tests locally everything is fine, and I can't seem to re-trigger the tests here for some reason.
> Screenshot 2019-07-17 at 13 50 58

You have to merge master into this branch and update snaps

PR #256, opened 2019-07-17
fix: set neutral button background to transparent

smithsamyan:


src/Molecules/Button/Button.tsx
```diff
@@ -34,7 +34,7 @@ const getBackgroundColor = (props: ThemedStyledProps<ButtonProps, Theme>) => {
   const { disabled, theme, variant, colorFn } = props;
 
   if (isNeutral(variant)) {
-    return ``;
+    return `transparent`;
```
mjmonline: `background-color: transparent;`

shmidt-i: Good catch!

shmidt-i:
Merged manually

PR #262, opened 2019-08-12
Feat/text input

shmidt-i:

PR #265, opened 2019-08-14
Feat/add bold option in select

Maxenboy:


jauniuse:
`optionWeight` prop is quite confusing, since it has to do something with label, not with option in expanded list.

and I think it's quite specific case which only you will use, I'd rather do this in the way that @shmidt-i did for `Input`, ie

https://github.com/nordnet/ui/pull/262/files#diff-9593e5242d8ac2e10b581fc576ab0c08R183

mjmonline:
@Maxenboy According to [Invision](https://nordnet.invisionapp.com/d/main#/console/16903964/358889175/preview) the dropdown component has no bold text. Would like to avoid the variations if possible. Has this been verified with designers?

PR #274, opened 2019-08-26
enable tracking in buttons and links

smithsamyan:
Added tracking context to enable tracking and implemented it in button and link


src/common/tracking.types.ts
```diff
@@ -0,0 +1,5 @@
+type TrackFunction = (componentName: string, e: React.SyntheticEvent, props: any) => void;
+
+export interface TrackingContextData {
```
shmidt-i: better `type` than interface, if you are not gonna extend it

PR #282, opened 2019-09-03
fix: reset background of button in link component

### LINKED ISSUE #279: Button looking like a link does not have transparent background in windows
image

Reproduced on latest chrome in windows 10 so not isolated to ie/edge

mjmonline:
Closes #279

### DIFF
--- a/src/Molecules/Link/Link.tsx
+++ b/src/Molecules/Link/Link.tsx
@@ -39,6 +39,7 @@ const StyledLink = styled(CleanLink)<LinkProps>`
 
 const StyledButton = styled(NormalizedElements.Button)<LinkProps>`
   ${p => getSharedStyle(p)}
+  background: none;
   cursor: ${p => (p.disabled ? 'not-allowed' : 'pointer')};
   border: none;
   border-radius: 0;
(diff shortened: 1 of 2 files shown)


src/Molecules/Link/Link.tsx
```diff
@@ -39,6 +39,7 @@ const StyledLink = styled(CleanLink)<LinkProps>`
 
 const StyledButton = styled(NormalizedElements.Button)<LinkProps>`
   ${p => getSharedStyle(p)}
+  background: none;
```
shmidt-i: I am thinking about adding this in my PR:
```css
  /* resetting button styles */
  border: none;
  margin: 0;
  padding: 0;
  width: auto;
  overflow: visible;

  background: transparent;

  /* Corrects font smoothing for webkit */
  -webkit-font-smoothing: inherit;
  -moz-osx-font-smoothing: inherit;

  /* Corrects inability to style clickable input types in iOS */
  -webkit-appearance: none !important;
  /* resetting button styles end */
```

mjmonline: Are you thinking in the NormalizedButton file? I believe the `font-smoothing` is outdated, the browsers have gotten updates since then which makes them render better without this rule. Also the `-webkit-appearance` is already defined in the NormalizedButton file.

PR #283, opened 2019-09-04
Add new RTMT icon

jonasjeeliasson:
Add new icon for RTMT. This icon requires a larger viewbox so this PR also enables setting a custom viewBox when needed. If not set the previously hard coded viewbox "0 0 24 24" will be used.


src/Atoms/Icon/IconBase.types.ts
```diff
@@ -17,6 +17,7 @@ export type StyledChildProps = {
 export type BaseProps = {
   className?: string;
   title?: string;
+  viewBox?: string;
```
mjmonline: This can be `viewBox?: '12' | '24' | '48'`. Don't think we want it to be totally custom. If you want to know why, please find me and lets discuss it

jonasjeeliasson: Sounds reasonable to lock it down to some allowed values, should it though rather be: `viewBox?: '0 0 12 12' | '0 0 24 24' | '0 0 48 48';`, or is that the same thing?

mjmonline: On second thought I think let's keep it as you have it, totally customizable

jonasjeeliasson: 👍

PR #284, opened 2019-09-04
feat: create access to inner components of modal

mjmonline:


src/Molecules/Modal/Modal.tsx
```diff
@@ -1,14 +1,26 @@
...
-  onClose: onCloseExternal,
-  ...rest
-}) => {
+const components = {
+  Dialog,
+};
+
+export const Modal: React.FC<Props> & {
+  /**
+   * This will allow you to customize
```
jesperwiner: nice documentation

PR #289, opened 2019-09-10
Feat: expose modal content

joel-nn:
Expose modal Content to be able to style it, only set max-height in desktop.


src/Molecules/Modal/ModalInner.tsx
```diff
@@ -52,8 +52,10 @@ const RightAlignedFlex = styled(Flexbox)`
   margin-left: auto;
 `;
 
-const Content = styled.div`
-  max-height: 50vh;
+export const Content = styled.div`
+  ${({ theme }) => theme.media.greaterThan(theme.breakpoints.md)} {
```
mjmonline: breakpoint.sm

PR #294, opened 2019-09-13
fix: controlled value

mjmonline:


src/Molecules/Input/Number/Number.tsx
```diff
@@ -153,24 +152,19 @@ const NumberInput: NumberComponent & {
     success,
     value: controlledValueRaw,
   } = props;
-  const [internalValue, setInternalValue] = useState(getNumberAsString(defaultValue));
-  const previousInternalValue = usePrevious(internalValue);
+  const [internalValue, setInternalValueState] = useState(getNumberAsString(defaultValue));
+  const setInternalValue = (val: string) => {
```
mjmonline: can we rename this function to `handleValueChange` and `setInternalValueState` back to `setInternalValue`?

jauniuse: jap, done

PR #295, opened 2019-09-13
Fix: add icons

Vikingjonsson:
Added new icons and renames alert to bell


src/Atoms/Icon/index.ts
```diff
@@ -1,15 +1,18 @@
-import { Alert } from './components/Alert';
 import { ArrowRight } from './components/ArrowRight';
 import { Attachement } from './components/Attachement';
 import { Bank } from './components/Bank';
+import { Bell } from './components/Bell';
 import { Bolt } from './components/Bolt';
+import { DiagonalWithCircle } from './components/DiagonalWithCircle';
```
mjmonline: alphabetical order please

src/Atoms/Icon/index.ts
```diff
@@ -30,18 +33,21 @@ import { WarningTriangle } from './components/WarningTriangle';
 import { Rtmt } from './components/Rtmt';
 
 export default {
-  Alert,
```
jauniuse: nah, this is breaking change. You should make it backwards compatible

jauniuse: if rename was really needed here, then export `Alert` also:

```js
export default {
  Alert: RenamedIcon,
  RenamedIcon,
}
```
and write a deprecation comment

mjmonline: he just put this in 5mins ago

jauniuse: sneaky

PR #296, opened 2019-09-14
make number input as type text

jauniuse:
input type number does not work well with localized values when using it together with react-intl. It's either we remove react-intl from this input or make it as text (as per this PR) with normalizing entered values to be numbers.

[inputmode](https://caniuse.com/#search=inputmode) is used for opening numerical keyboard on mobile devices when one has input type text, it's supported in few latest versions of  chrome and iOS safari which covers most of the users

### DIFF
--- a/src/Molecules/Input/Number/Number.tsx
+++ b/src/Molecules/Input/Number/Number.tsx
@@ -1,6 +1,7 @@
 import React, { useState, useRef } from 'react';
 import { injectIntl } from 'react-intl';
 import styled, { css } from 'styled-components';
+import * as R from 'ramda';
 import { Props, NumberComponent } from './Number.types';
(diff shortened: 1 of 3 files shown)


src/Molecules/Input/Number/Number.tsx
```diff
@@ -102,10 +104,6 @@ const Input = styled(NormalizedElements.Input).attrs({ type: 'number' })<Partial
     -webkit-appearance: none; /* stylelint-disable-line property-no-vendor-prefix */
```
mjmonline: This rule needs to be removed, the whole block, with the margin aswell

jauniuse: done

PR #298, opened 2019-09-17
Feat/labeled value with tertiary title

joel-nn:
Optionally make the label in labeled value tertiary, I accidentally named the branch secondary. Sorry about that.


src/Molecules/LabeledValue/LabeledValue.types.ts
```diff
@@ -4,6 +4,7 @@ import { FormattedMessage } from 'react-intl';
 export type Props = {
   label: React.ReactText | FormattedMessage.MessageDescriptor;
```
mjmonline: probably it's better to let label take a `React.ReactNode` as well. Then we dont have to specify all the current and future variations of `labelType`. Look at `PageHeaderCard` how that is done

joel-nn: I thought we wanted to be strict on purpose to avoid snowflaking?

src/Molecules/LabeledValue/LabeledValue.types.ts
```diff
@@ -1,9 +1,9 @@
 import React, { ReactNode } from 'react';
-import { FormattedMessage } from 'react-intl';
 
 export type Props = {
-  label: React.ReactText | FormattedMessage.MessageDescriptor;
+  label: string | React.ReactNode;
   children: ReactNode;
+  labelType?: 'tertiary';
```
mjmonline: this is no longer needed right?

src/Molecules/LabeledValue/LabeledValue.types.ts
mjmonline: the first solution was kinda a slowflake as well, at least this is more consistant

src/Molecules/LabeledValue/LabeledValue.types.ts
joel-nn: nope

PR #299, opened 2019-09-18
Feat: input checkbox

mjmonline:


mjmonline:
I will rename the internal `FormField` component found in `Input > FormField`. Just didn't want to do it in this PR because it's already big as is

PR #303, opened 2019-09-19
feat: 🎸 add error prop to select component

crewneck:
Added error prop to the select component to handle validation when the user has not selected an option in the select box.


mjmonline:
@crewneck I have a PR coming up which will let you wrap your `<Select>` in a `<FormField>` and that component will be able to display error. So please hold off on this

mjmonline:
@crewneck You can now use `FormField` to achive this. You can also look at the story for `Input.Number` to see how a input is used in combination with `FormField`

PR #304, opened 2019-09-19
feat: move out form field component

mjmonline:


Maxenboy:
The required label doesn't seem to work for the number input in storybook 
Screen Shot 2019-09-19 at 14 16 03

mjmonline:
> The required label doesn't seem to work for the number input in storybook
> Screen Shot 2019-09-19 at 14 18 00
> Seem like `showRequired` is missing in the story I guess
> but `showRequired` is not stated in the Formfield required story, me confused :S

Updated to show the star. You're looking in the wrong `FormField` the one under `Input` is an internal only component that we will change the name of. The correct component is under `Molecules > FormField`

src/Molecules/FormField/FormField.types.ts
```diff
@@ -3,13 +3,16 @@ export type Props = {
   label: string;
   /** But you can hide it visually */
   hideLabel?: boolean;
+
```
KitsuneKelso: Just curious, why the extra newlines?

mjmonline: That's just my OCD :)

PR #313, opened 2019-09-25
Fix: input checkbox refactor

mjmonline:


shmidt-i:
https://nordnet-ui-git-fix-input-checkbox-refactor.nordnet1.now.sh/?path=/story/molecules-input-checkbox--in-a-group-with-error - should be without star by default

PR #321, opened 2019-09-30
fix: less padding bottom on cardwithtabs

mjmonline:


shmidt-i:
Are we sure, nobody is snowflaking with this component in ui-mutations @mjmonline ?

mjmonline:
> Are we sure that nobody is snowflaking with this component in ui-mutations @mjmonline ?

Didn't find anything

PR #349, opened 2019-10-08
fix: support full width for select

mjmonline:


src/Molecules/Input/Select/Select.tsx
```diff
@@ -165,7 +165,7 @@ const Select = (props: Props) => {
         value: isControlledMode ? valueFromProps : _state.value,
       },
     });
-  }, [options, placeholder, isControlledMode, valueFromProps, dispatch]);
+  }, [options, placeholder, isControlledMode, valueFromProps, dispatch, _state.value]);
```
jauniuse: is `_state.value` addition here deliberate?

mjmonline: good catch, fixed

mjmonline: good catch, fixed

PR #350, opened 2019-10-08
feat: bump react and react-router deps

Dhq:


package.json
```diff
@@ -66,8 +66,8 @@
     "react-focus-lock": "^2.0.3",
     "react-intl": "^2.8.0",
     "react-remove-scroll": "^2.0.4",
-    "react-router": "^4.3.1",
-    "react-router-dom": "^4.3.1",
+    "react-router": "^5.1.2",
```
shmidt-i: can you please make it `5.1.2` here instead of `^5.1.2`? I know we had it with cap for 4.3.1, but anyway :)

Dhq: Sure!

PR #355, opened 2019-10-10
feat: 🎸 left align number when no steppers for input.number

jonasjeeliasson:
Made the number left aligned when steppers aren't used. Steppers aren't used when noSteppers is set or when either/both left and right addon is used.  

no_steppers
right_addon
left_addon
both_addons
steppers

### DIFF
--- a/src/Molecules/Input/Number/Number.tsx
+++ b/src/Molecules/Input/Number/Number.tsx
@@ -109,7 +109,7 @@ const Input = styled(NormalizedElements.Input).attrs({ type: 'text' })<Partial<P
       ? `
       padding-top: ${p.theme.spacing.unit(2)}px;
       padding-bottom: ${p.theme.spacing.unit(2)}px;
-      padding-left: ${p.leftAddon ? p.theme.spacing.unit(10) : p.theme.spacing.unit(2)}px;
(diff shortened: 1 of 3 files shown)


src/Molecules/Input/Number/Number.tsx
```diff
@@ -102,12 +102,15 @@ const Input = styled(NormalizedElements.Input).attrs({ type: 'text' })<Partial<P
...
-  text-align: center;
+  text-align: ${p => (p.showSteppers ? 'center' : 'left')};
   box-sizing: border-box;
   ${p =>
     p.leftAddon || p.rightAddon
       ? `
-      padding: ${p.theme.spacing.unit(2)}px ${p.theme.spacing.unit(10)}px;
+      padding-top: ${p.theme.spacing.unit(2)}px;
+      padding-bottom: ${p.theme.spacing.unit(2)}px;
+      padding-left: ${p.leftAddon ? p.theme.spacing.unit(10) : p.theme.spacing.unit(2)}px;
```
mjmonline: Input.Text has 8 units of left padding, let's keep it consistent

jonasjeeliasson: Ah, should have double check that of course :) Will update it.

PR #368, opened 2019-10-16
feat: add minimize and maximize icons

jesperwiner:


jesperwiner:
> Try not to name the icons after what they will do rather name them for what they are

Well, any suggestion is welcome

PR #375, opened 2019-10-21
feat: 🎸 add ie check as util function and hook

smithsamyan:


src/common/utils.ts
```diff
@@ -59,3 +59,6 @@ const isInvalid = R.anyPass([
 ]);
 
 export const isValidDateTimeNumber = R.complement(isInvalid) as (x: any) => x is number;
+
+export const isIE = (): boolean =>
+  window.matchMedia('all and (-ms-high-contrast: none), (-ms-high-contrast: active) ').matches;
```
iamstarkov: this will explode on the server, please change

shmidt-i:
@smithsamyan please update the PR

smithsamyan:
The current solution will explode on server as @iamstarkov pointed out above. As I see it now, there are three ways to go about this

1. Make it work for both server and client using userAgent sniffing. This can be very [unreliable](https://blogs.msdn.microsoft.com/ieinternals/2013/09/21/internet-explorer-11s-many-user-agent-strings/) 
2. Make it return false on the server but use it on client. This may cause a flicker between the SSR version and the rehydrated version on IE11
3. Scrap and instead let the developers use feature detection for their js or css in their own code

@shmidt-i What do you think, how should we proceed?

iamstarkov:
I vote for the 3rd option

shmidt-i:
@smithsamyan let's discuss it tomorrow one more time and make a final decision on this one

mjmonline:
Can't we try to find a solution where we put IE specific code in  html tag?

smithsamyan:
Hmm maybe we could instead do like Media component does and on server with different css so it's what is IE specific is not displayed on client and vice versa.

shmidt-i:
@smithsamyan look into option Mansour suggested. Will close this one for now

PR #376, opened 2019-10-21
feat: 🎸 implement slider

crewneck:


shmidt-i:
Nice job, but yeah, we need to add more polishing and change the name probably. Start with the mutations if you need it right away

PR #377, opened 2019-10-21
feat: add shareville logo

jesperwiner:


src/Atoms/Icon/components/SharevilleLogo.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { BaseProps } from '../IconBase.types';
+
+export const SharevilleLogo = (props: BaseProps) => {
+  return (
+    <IconBase viewBox="0 0 166.54 203.5" {...props} title="Shareville">
+      <defs>
+        <style
+          dangerouslySetInnerHTML={{
+            __html:
+              '.svLogoGradientClass-1{fill:url(#linear-gradient);}.svLogoGradientClass-2{fill:url(#linear-gradient-2);}.svLogoGradientClass-3{fill:url(#linear-gradient-3);}',
```
shmidt-i: • Need to check duplication of id problem on chrome (we had one for spinner https://nordnet.github.io/ui/?path=/story/atoms-spinner--regression-2-spinners-with-same-color-and-size-affects-each-other-in-chrome)
• Need to check in different browsers

Will do it tomorrow, no rush?

jesperwiner: yes, no rush, I just copied the one from old-ui-kit. Tried to SVGO it but that seems to fail due to the style tag.

PR #378, opened 2019-10-21
Add optional footer to Modal

anneback:
Adds a new optional prop `footer` to `Modal`.

### DIFF
--- a/src/Molecules/Modal/Modal.stories.tsx
+++ b/src/Molecules/Modal/Modal.stories.tsx
@@ -85,7 +85,7 @@ export const footerStory = () => {
     const footer = (
       <div>
         <Typography type="primary" as="p">
-          This is a modal which is a ReactNode, which could e.g. contain a button
+          This is a footer which is a ReactNode, which could e.g. contain a button or only text
         </Typography>
         <button type="button" onClick={() => {}}>
           Button
--- a/src/Molecules/Modal/Modal.stories.tsx
+++ b/src/Molecules/Modal/Modal.stories.tsx
@@ -70,6 +70,72 @@ defaultStory.story = {
   name: 'Default',
 };
 
+export const footerStory = () => {
+  const Example = () => {
+    const [open, setOpen] = useState(true);
+
+    const onOpen = () => {
+      setOpen(true);
+    };
+
+    const onClose = () => {
+      setOpen(false);
+    };
+
+    const footer = (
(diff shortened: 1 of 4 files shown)


shmidt-i:
Interesting approach, I like it! That would make a lot cases, namely button groups at modal bootom easier to implement.

src/Molecules/Modal/ModalInner.tsx
```diff
@@ -48,6 +48,10 @@ const Button = styled(NormalizedElements.Button)`
   cursor: pointer;
 `;
 
+const Footer = styled(Flexbox)`
```
shmidt-i: Can it be just `<Box pt={4} />`?

src/Molecules/Modal/Modal.stories.tsx
```diff
@@ -70,6 +70,72 @@ defaultStory.story = {
...
+    };
+
+    const onClose = () => {
+      setOpen(false);
+    };
+
+    const footer = (
+      <div>
+        <Typography type="primary" as="p">
+          This is a modal which is a ReactNode, which could e.g. contain a button
```
KitsuneKelso: Maybe this should say that "this is a `footer`" instead of a `modal`, but that's a big nitpick. :)

src/Molecules/Modal/ModalInner.tsx
shmidt-i: Also, can we bake in fix that we had for modals regarding buttons being at the bottom and outline being truncated because of that?
image

Fix is adding 1px padding and removing it with -1px negative margin, can be part of the same box
e.g. `<Box p="1px" m="-1px" pt={4} />`

src/Molecules/Modal/Modal.stories.tsx
anneback: updated documentation

src/Molecules/Modal/ModalInner.tsx
anneback: I'll squeeze it in!

shmidt-i: Actually, maybe it won't be enough. Original solution was 
```
${Modal.components.Content} {
    margin: -${p => p.theme.spacing.unit(1)}px;
    padding: ${p => p.theme.spacing.unit(1)}px;
  }
```

PR #387, opened 2019-10-24
feat: 🎸 add placeholder to Input.Number

jonasjeeliasson:
We need a placeholder for the Input.Number.


src/Molecules/Input/Number/Number.tsx
```diff
@@ -8,6 +8,7 @@ import NormalizedElements from '../../../common/NormalizedElements';
 import { getStringAsNumber, getNumberAsString } from './utils';
 import { isNumber, isString, isUndefined, assert } from '../../../common/utils';
 import adjustValue from './adjustValue';
+import { placeholderNormalizaion } from '../Text/Text';
```
jonasjeeliasson: I reused the placeholderNormalization from Text (the same way is used in TextArea)

jonasjeeliasson:
Hmm got some conflicts, will have to check them tomorrow.

jonasjeeliasson:
> Hmm got some conflicts, will have to check them tomorrow.

mjmonline:
How does the use case look? Also if there is a valid use case then include a story of it

jonasjeeliasson:
Hi,

The use case is that we need placeholder saying "Belopp". For all except
one of our number inputs, we want this placeholder.

Do you mean I should include a screenshot here displaying its actual use
case or that I should change the story in storybook with a real use case
instead of the generic one with the text placeholder which I've already
added?

shmidt-i:
@jonasjeeliasson sry for late answer, could you please change placeholder to only appear when there is `noSteppers`? that's the only thing I didn't like couple of days ago, but then totally forgot about this ;)

jonasjeeliasson:
@shmidt-i Now the placeholder won't be visible with steppers. Added multiple cases in the storybook to show of how it looks with different addons and such.

PR #390, opened 2019-10-29
Feat: faded scroll

mjmonline:


src/Atoms/FadedScroll/FadedScroll.mdx
```diff
@@ -0,0 +1,15 @@
...
+import { FadedScroll } from './FadedScroll.tsx';
+
+# FadedScroll
+
+This component provides a faded look to the bottom of a vertically scrollable element.
+
+## Usage
+
+This component is only intended to give an effect on screen sizes larger than 'md'. We try to not have scrollbars in mobile.
+You can provide a maxHeightDesktop to the component or let this component take all available space of parent and if the content exceeds that hight then show the scrollbar.
```
shmidt-i: `You can provide a maxHeightDesktop to the component or let this component take all available space of parent and if the content exceeds that hight then show the scrollbar.` - this part should also be in `maxHeightDesktop` prop's JSDoc.

Also, nothing about flex parent, is it still needed?
Also, looking at stories, fadeHeight is understandable, but difference between other two is hard to grasp

src/Atoms/FadedScroll/FadedScroll.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+);
+
+export const defaultStory = () => <FadedScroll maxHeightDesktop={40}>{content}</FadedScroll>;
+
+defaultStory.story = {
+  name: 'Default',
+};
+
+export const withHeightContainedByParent = () => (
+  <ContainingParent>
```
shmidt-i: Maybe here better to use `<div style={{height: '200px'}}>`, just so people don't need to switch contexts

src/Atoms/FadedScroll/FadedScroll.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <FadedScroll>{content}</FadedScroll>
+  </ContainingParent>
+);
+
+withHeightContainedByParent.story = {
+  name: 'With height contained by parent',
```
shmidt-i: `With height declared by parent` maybe?

src/Atoms/FadedScroll/FadedScroll.tsx
```diff
@@ -0,0 +1,70 @@
...
+    background: linear-gradient(0deg, rgba(255, 255, 255, 1) 0%, rgba(255, 255, 255, 0) 100%);
+    opacity: 1;
+    transition: 0.3s opacity ease-out;
+
+    ${p => p.intersectionOnScreen && `opacity: 0;`}
+  }
+`;
+
+const Inner = styled.div<Pick<Props, 'maxHeightDesktop'>>`
+  ${p => p.theme.media.greaterThan(p.theme.breakpoints.md)} {
```
shmidt-i: Are we sure, that this is needed only in desktop? I totally see SearchModal on the mobile where you have a big scrollable list of items and better to have scroll there

src/Atoms/FadedScroll/FadedScroll.types.ts
```diff
@@ -0,0 +1,10 @@
+export type InternalProps = {
+  intersectionOnScreen?: boolean;
+};
+
+export type Props = {
+  maxHeightDesktop?: string | number;
+  fadeHeight?: string | number;
+};
+
+export type Component = React.FC<Props & InternalProps>;
```
shmidt-i: Why expose `InternalProps`? 
`intersectionOnScreen` is currently showing in `Props` in docs. But it's not propagating to Outer. Some fix needed :)

src/Molecules/CardWithTitle/CardWithTitle.tsx
```diff
@@ -2,12 +2,24 @@ import React from 'react';
...
+import { CardWithTitleComponent, Props } from './CardWithTitle.types';
 
 const StyledHeader = styled.header`
   padding: ${({ theme }) => theme.spacing.unit(5)}px;
 `;
 
+type ScrollableProp = Pick<Props, 'scrollable'>;
+const StyledCard = styled(Card)<ScrollableProp>`
+  ${p =>
+    p.scrollable &&
```
shmidt-i: But this without `<FadedScroll>` inside doesn't make any sense, right?

src/common/Hooks/useOnScreen.tsx
```diff
@@ -0,0 +1,34 @@
+import { useState, useEffect } from 'react';
+
+const useOnScreen = (ref: React.RefObject<any>, rootMargin: string = '0px') => {
```
shmidt-i: Some jsdoc here, explaining what hook does

shmidt-i: Also, what are returning values and what they mean

shmidt-i:
Also, check if we can move <ShowMore> from mutations using this component

src/Molecules/CardWithTitle/CardWithTitle.tsx
mjmonline: The card requires these styles if `<FadedScroll>` has no height on it. Maybe we should solve this in another way, have any ideas?

src/Atoms/FadedScroll/FadedScroll.tsx
mjmonline: We are trying to avoid scrollbars on mobile because most likely the page will also have a scrollbar which means we will have double scrollbars. Maybe we should enable it with a prop like `enableMobileFade`

src/Atoms/FadedScroll/FadedScroll.stories.tsx
mjmonline: True, but I forgot to wrap the height style in a media query so can't do this anymore for this story. But will do this for the story with `enableMobileFade`

src/Atoms/FadedScroll/FadedScroll.mdx
shmidt-i: `hight` -> `height` typo

src/Molecules/CardWithTitle/CardWithTitle.md
```diff
@@ -2,5 +2,9 @@
...
 The title that you pass in can be anything, can be custom component as well.
-________________
-In the future we may incorporate some more styling here, but for now you need to create mutations out of this.
\ No newline at end of file
+
+---
+
+In the future we may incorporate some more styling here, but for now you need to create mutations out of this.
+
+❗️Important, if you are want to use `FadedScroll` without it's `maxHeight` prop inside this component which means that you want Flexbox to just give the content area all the available space thats left. Then wrap all content in a container element with the following styles: `height: 100%; padding-bottom: flex-grow: 1; min-height: 0; /* Firefox fix */`
```
shmidt-i: `padding-bottom: flex-grow: 1;`  feels like a typo :)

src/Molecules/CardWithTitle/CardWithTitle.tsx
shmidt-i: Let's discuss it tomorrow

src/common/Hooks/useOnScreen.tsx
```diff
@@ -0,0 +1,39 @@
+import { useState, useEffect } from 'react';
+
+/**
+ * This hook is used to check if passed ref is visible on the screen.
+ * This is useful if you for example want to know when your element is scrolled into view.
+ * The hook will return true or false depending on if the ref is visible on the screen or not.
```
shmidt-i: Add some stuff about `null` as returning value, which is IE11 or SSR

src/Atoms/FadedScroll/FadedScroll.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+    {content}
+  </FadedScroll>
+);
+
+withHeightOfFadeChanged.story = {
+  name: 'With height of fade changed',
+};
+
+export const withMobileFadeEnabled = () => (
+  <div style={{ height: '220px' }}>
```
shmidt-i: Let's use https://www.npmjs.com/package/@storybook/addon-viewport?

shmidt-i: Can do in a separate PR after

PR #392, opened 2019-10-29
refactor(number): make variable naming better

shmidt-i:


shmidt-i:
Gonna address it in a separate PR

PR #397, opened 2019-11-04
Fix/viewbox for plus and minus

joel-nn:
Added  viewBox to new plus and minus


shmidt-i:
image
image
not aligned within viewbox, one px off probably :)

PR #405, opened 2019-11-06
feat: 🎸 add no ssr component

smithsamyan:
Add a component to not render something on server

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "color": "^3.1.1",
     "focus-within-polyfill": "^5.0.4",
     "framer-motion": "^1.6.2",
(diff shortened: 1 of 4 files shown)


mjmonline:
We use [use-ssr](https://github.com/alex-cory/use-ssr) to do these kind of things. Do you prefer the component in this PR?

smithsamyan:
> We use [use-ssr](https://github.com/alex-cory/use-ssr) to do these kind of things. Do you prefer the component in this PR?

Ohh.. I asked @shmidt-i If we had any standard way and then suggested this component. I will just use that instead then. Closing

PR #417, opened 2019-11-15
Fix/button sync

mjmonline:


src/Molecules/Button/Button.tsx
```diff
@@ -1,206 +1,62 @@
...
-    return `color: ${isSecondary(variant) ? color || theme.color.cta : theme.color.buttonText};`;
-  };
-
-  const getBorderColor = () => {
-    if (disabled || isNeutral(variant)) {
-      return 'transparent';
+const StyledButton = styled(NormalizedElements.Button)<ButtonProps>`
+  ${p => {
+    if (isSecondary(p.variant)) {
+      return secondaryStyles;
```
shmidt-i: Does it work like that?

```
${p => css`
  p => p.something
`}
```

src/Molecules/Button/Button.styles.tsx
```diff
@@ -0,0 +1,134 @@
...
+  ${p => (p.fullWidth ? `display: flex; width: 100%;` : `display: inline-flex;`)}
+`;
+
+const minHeight = css<ButtonProps>`
+  min-height: ${p => p.theme.spacing.unit(HEIGHT[p.size!])}px;
+`;
+
+const padding = css<ButtonProps>`
+  ${p => {
+    const unit = PADDING_HORIZONTAL[p.size!];
```
shmidt-i: why are we sure that size is presented here?

src/Molecules/Button/Button.styles.tsx
```diff
@@ -0,0 +1,134 @@
...
+const shared = css<ButtonProps>`
+  position: relative;
+  box-sizing: border-box;
+  align-items: center;
+  justify-content: center;
+  ${p => (p.fullWidth ? `display: flex; width: 100%;` : `display: inline-flex;`)}
+`;
+
+const minHeight = css<ButtonProps>`
+  min-height: ${p => p.theme.spacing.unit(HEIGHT[p.size!])}px;
```
shmidt-i: `!` has a different semantics in typescript, like when you are saying "I've checked already and it's not undefined, I'm sure".

Here you have different situation: the props that you are providing to the `StyledButton` are not the same as the props provided to the `Button` component, you already applying defaults, e.g. for the `size` prop. 
So more correct way would be to specify here that it's not ButtonProps, but rather some derivative with `size` being not required.

For example it can be 
```
type InnerProps = {
  [K in keyof ButtonProps]: ButtonProps[K];
  size?: ButtonProps['size']; // adding ? here
}
```

src/Molecules/Button/Button.tsx
mjmonline: yepp, no problems

shmidt-i: need to fix :)

PR #423, opened 2019-11-22
Feat/button loading state

mjmonline:

PR #425, opened 2019-11-25
Fix/button size small cleanup

mjmonline:


mjmonline:
Please look at these PR's before: https://github.com/nordnet/ui/pull/417 https://github.com/nordnet/ui/pull/423

PR #429, opened 2019-12-03
fix: add type to link that looks like button

jesperwiner:


jesperwiner:
Please merge and build asap

src/Molecules/Input/Select/__snapshots__/Select.stories.storyshot
```diff
@@ -5162,6 +5162,7 @@ Array [
             <div
               className="c7"
               onClick={[Function]}
+              type="button"
```
shmidt-i: @mjmonline leaked

PR #430, opened 2019-12-05
Feat/drawer

mjmonline:


src/Molecules/Drawer/Drawer.types.ts
```diff
@@ -0,0 +1,8 @@
+export type Props = {
+  className?: string;
+  onClose?: Function;
+  title?: string;
```
shmidt-i: ReactNode?

src/Molecules/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,127 @@
...
+    x: 70,
+  },
+  animate: {
+    opacity: 1,
+    x: 0,
+  },
+  transition: {
+    type: 'spring',
+    damping: 15,
+    stiffness: 200,
```
shmidt-i: I think these needs to be adjusted a bit, too much time wiggling after it's appeared

shmidt-i: Also, overall timing can be smaller

mjmonline: We can take a look at this later, this is the same values for the `<Modal />`

yarn.lock
```diff
@@ -1646,15 +1646,15 @@
     lodash "^4.17.4"
 
 "@semantic-release/commit-analyzer@^7.0.0-beta.1":
-  version "7.0.0-beta-.2"
-  resolved "https://registry.yarnpkg.com/@semantic-release/commit-analyzer/-/commit-analyzer-7.0.0-beta-.2.tgz#5f30919eabee8f876c5f80d270274dca34933f76"
-  integrity sha512-tburFNeooqb8WrqTfkXroUpvb1jbRP15xugOcWdvnNOlDLHL5D//8BK2Q6TOY+G3RhrL1qaZIuoQ7paqia3x1A==
+  version "7.0.0-beta.7"
+  resolved "https://registry.yarnpkg.com/@semantic-release/commit-analyzer/-/commit-analyzer-7.0.0-beta.7.tgz#d5b66ed41c0b774783bbaa6132189c6c84e34f0b"
+  integrity sha512-1mhphBfra8n4HZpIctrqbA8l7fXbBJ7lj2wgEfzL7kijAj4f6MUiRA9JmPAweRAnTyx7H9+FAQ7t8cE/7jsPJg==
```
mjmonline: Had to update yarn lock, this was resolving to a url that was giving a 404. Causing now to fail it's build

PR #437, opened 2019-12-19
Fix/modal autofocus

mjmonline:


src/Molecules/Modal/ModalInner.tsx
```diff
@@ -100,35 +112,28 @@ export const ModalInner: React.FC<InnerProps> = ({
...
-                  )}
-                  {!hideClose && (
-                    <CloseButton type="button" onClick={onClose}>
-                      <Icon.CrossThin size={5} title="Close this dialog" />
-                    </CloseButton>
-                  )}
-                </Flexbox>
-              </Box>
+            <Dialog role="dialog" show={show} aria-labelledby={titleId} {...animationProps}>
+              <Header>
```
shmidt-i: if there is `hideClose` and no `title` presented, then we should remove `Header`

PR #452, opened 2020-01-31
Overlay tooltip

Vikingjonsson:
z-index for tooltips on overlays

### DIFF
--- a/src/Molecules/Tooltip/Tooltip.tsx
+++ b/src/Molecules/Tooltip/Tooltip.tsx
@@ -73,10 +73,6 @@ export const Tooltip: TooltipComponent = ({
   const { isVisible, triggerRect } = tooltip;
   const tooltipPosition = getToolTipPosition(position);
 
-  console.log('Tooltip:', tooltip);
-  console.log('Trigger:', trigger);
-  console.log('Children:', children);
-
   return (
     <>
       {cloneElement(children, trigger)}
--- a/src/Molecules/Tooltip/Tooltip.tsx
+++ b/src/Molecules/Tooltip/Tooltip.tsx
@@ -1,5 +1,6 @@
 import React, { cloneElement } from 'react';
 import styled from 'styled-components';
+import * as R from 'ramda';
 import { useTooltip, TooltipPopup } from '@reach/tooltip';
 import { TooltipComponent, Props } from './Tooltip.types';
 import { Typography } from '../..';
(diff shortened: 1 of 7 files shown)


src/Molecules/Tooltip/Tooltip.tsx
```diff
@@ -7,8 +7,10 @@ import Triangle from './Triangle';
 import { leftAfter, leftCenter, leftBefore, topCenter, topOver, topUnder } from './utils';
 import { BORDER_SIZE } from './consts';
 
-const StyledTooltip = styled(TooltipPopup)`
-  z-index: 1;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
jauniuse: is this needed?

Vikingjonsson: it will give a squiggle line otherwise telling me that the boolean is not used

src/theme/theme.types.ts
```diff
@@ -271,6 +271,7 @@ export type Theme = {
     header: 200;
     dropdown: 300;
     overlay: 400;
+    overlayTooltip: 401;
```
jauniuse: maybe better to keep pattern and have `overlayTooltip` as 500 and increase `modal` to 600

src/Molecules/Tooltip/Tooltip.tsx
jauniuse: ah, you can do

```js
(props) => <TooltipPopup {...R.omit(['overlayTooltip'], props)}/>
```

src/Molecules/Tooltip/Triangle/Triangle.tsx
```diff
@@ -84,13 +84,13 @@ const Arrow = styled.div<InternalArrowProps>`
...
   &::before {
-    z-index: 1;
+    z-index: ${p => (p.overlayTooltip ? ({ theme }) => theme.zIndex.overlayTooltip : 1)};
     left: 0;
     top: 0;
   }
 
   &::after {
-    z-index: 2;
+    z-index: ${p => (p.overlayTooltip ? ({ theme }) => theme.zIndex.overlayTooltip + 1 : 2)};
```
jauniuse: aghm, wouldnt it be better to make these all as constants in `theme.zIndex` ?

Vikingjonsson: thought this would make less errors if changed in the future, maybe not

src/Molecules/Tooltip/Tooltip.tsx
Vikingjonsson: i like that one, thank you :D

Vikingjonsson:
made a foreground version now

PR #457, opened 2020-02-06
fix: make modal header and footer not grow

mjmonline:


mjmonline:
@shmidt-i lets hope not 😜

PR #463, opened 2020-02-14
fix: enable mobile fade in drawer

Vikingjonsson:


shmidt-i:
@mjmonline didn't we have it like this before? :)

src/Molecules/Drawer/Drawer.tsx
```diff
@@ -127,7 +127,7 @@ export const Drawer = (React.forwardRef<HTMLDivElement, Props>(
                   <Icon.CrossThin size={CROSS_SIZE} title="Close this drawer" />
                 </CloseButton>
               </TitleWrapper>
-              <FadedScroll>{children}</FadedScroll>
+              <FadedScroll enableMobileFade>{children}</FadedScroll>
```
shmidt-i: @mjmonline didn't we have it like this before? :)

mjmonline: No, but this component should in my opinion not be baked in with the Drawer by default. The Modal component for example just takes children and there you can choose to have a FadedScroll or not

Vikingjonsson: yes that sounds like a solution for this

PR #470, opened 2020-02-24
feat: 🎸 additional input text size options

crewneck:
Added additional size to be able to match height with Input.Select

### DIFF
--- a/src/Molecules/Input/Text/Text.tsx
+++ b/src/Molecules/Input/Text/Text.tsx
@@ -7,8 +7,19 @@ import NormalizedElements from '../../../common/NormalizedElements';
 
 const hasError = (error?: Props['error']) => error && error !== '';
 
+const getInputHeight = (size: string | undefined) => {
+  switch (size) {
+    case 's':
+      return 8;
+    case 'm':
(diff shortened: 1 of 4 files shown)


src/Molecules/Input/Text/Text.tsx
```diff
@@ -7,8 +7,19 @@ import NormalizedElements from '../../../common/NormalizedElements';
 
 const hasError = (error?: Props['error']) => error && error !== '';
 
+const getInputHeight = (size: string | undefined) => {
+  switch (size) {
+    case 's':
+      return 8;
+    case 'm':
+      return 10.75;
```
jauniuse: why?

jauniuse: `Input.Select` has 8 for small and 10 for medium, so it should be same here:
https://github.com/nordnet/ui/blob/master/src/Molecules/Input/Select/lib/wrappers/FormFieldOrFragment.tsx#L24
 I believe you just need to update type to include `m`, like here:
https://github.com/nordnet/ui/blob/master/src/Molecules/Input/Select/Select.types.ts

PR #472, opened 2020-02-24
feat: 🎸 add some onMouse props to Button component

robineng:
I'm very unsure about the best way to solve this problem but basically I want to allow some (or preferably all) React button attributes on the UI button component. Allowing all of them does however pose some problems as some conflict with the current Button props.

I've also thought about spreading rest of the props into the sharedProps object instead of picking the three I want but that would require me to omit some other props so idk.

### DIFF
--- a/src/Molecules/Button/Button.types.ts
+++ b/src/Molecules/Button/Button.types.ts
@@ -4,6 +4,7 @@ import { Theme } from '../../theme/theme.types';
 
 type Colors = Theme['color'];
 type ColorFn = (t: Theme) => Colors['cta'] | Colors['negative'];
+
 export type ButtonProps = {
   /** @default primary */
   variant?: 'primary' | 'secondary' | 'neutral';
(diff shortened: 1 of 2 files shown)


src/Molecules/Button/Button.types.ts
```diff
@@ -29,7 +30,7 @@ export type ButtonProps = {
   as?: any;
   loading?: boolean;
   ref?: React.Ref<HTMLAnchorElement> | React.Ref<HTMLButtonElement>;
-};
+} & Pick<React.DOMAttributes<HTMLButtonElement>, 'onMouseEnter' | 'onMouseLeave' | 'onMouseOver'>;
```
shmidt-i: The problem with this approach though is that we can have either `HTMLAnchorElement` or `HTMLButtonElement`, so if you will try to use  onMouseEnter on anchor, you might have some issues :)

shmidt-i: Maybe
```typescript
& Pick<React.DOMAttributes<HTMLButtonElement | HTMLAnchorElement>, 'onMouseEnter' | 'onMouseLeave' | 'onMouseOver'>;
```
will do

robineng: Fixed 👍

PR #487, opened 2020-03-20
feat: 🎸 add breakpoint props to CssGrid.Item

robineng:


src/Atoms/CssGrid/CssGrid.tsx
```diff
@@ -7,7 +7,7 @@ import {
   Gutter,
   TemplateColumn,
   TemplateRow,
-  AreaName,
+  ItemProps,
```
iamstarkov: is it a breaking change?

robineng: No I don't think so. The types are never exported and I've only added props to the CssGrid.Item component

PR #490, opened 2020-03-24
Rating star colors

Vikingjonsson:


src/theme/theme.types.ts
```diff
@@ -131,10 +131,14 @@ export type Theme = {
     spinnerWhite: RawColor['white'];
     /** index */
     streamingBolt: RawColor['index'];
+    /** index */
+    ratingStar: RawColor['index'];
```
mjmonline: There is some gray rating icons coming up, maybe this should be `ratingStarColored`?

PR #502, opened 2020-04-07
Feat/add thin arrow

nn-davsko:
new arrow icon, takes direction as props

### DIFF
--- a/src/Atoms/Icon/Icon.stories.tsx
+++ b/src/Atoms/Icon/Icon.stories.tsx
@@ -20,9 +20,27 @@ export const inlineStory = () => (
   </span>
 );
 
-export const iconWithProps = () => (
+export const thinArrowUp = () => (
   <span>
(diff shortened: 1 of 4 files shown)


src/Atoms/Icon/Icon.stories.tsx
```diff
@@ -18,6 +20,12 @@ export const inlineStory = () => (
   </span>
 );
 
+export const iconWithProps = () => (
+  <span>
+    <Icon.ThinArrow direction={select('Direction', ['up', 'right', 'down', 'left'], 'up')} />
+  </span>
+);
+
```
mjmonline, line 28: Because we use these as tests and in the future possibly as screenshot tests a story for each mode would be good

nn-davsko: Ok got it, I'll add on for each.

PR #504, opened 2020-04-08
fix: change types so it accepts react nodes

smithsamyan:
Changes types to accept react nodes. This works today already but the types complain. String is a type of ReactNode so the current implementations will keep working with the correct types. This is also how CardWithTitle does it

### DIFF
--- a/src/Molecules/FormField/FormField.types.ts
+++ b/src/Molecules/FormField/FormField.types.ts
@@ -4,12 +4,12 @@ export type LabelAddonProp = {
 };
 
 export type Props = {
-  label?: string;
+  label?: React.ReactNode;
   hideLabel?: boolean;
   children: React.ReactNode;
   className?: string;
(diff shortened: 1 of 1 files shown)


smithsamyan:
@mjmonline @jesperwiner It turns out label doesn't work because of `labelText` converting label into string. I was wondering If we should skip optional addon when a user sends in a component and just render it instead so they have more control over how it works?

mjmonline:
@smithsamyan Well, what are you trying to do? I would like to have a consistent way of handling required look

smithsamyan:
@mjmonline I'm trying to allow extraInfo to be a component rather than just text. Currently we're adding a visually hidden extraInfo for the screen readers but for other users we would like to style it differently and put it in a different place.

I thought maybe this was needed for all these string types but maybe we should just add it for extraInfo?

PR #505, opened 2020-04-14
Risk level

Vikingjonsson:


src/Molecules/Risk/Risk.mdx
```diff
@@ -0,0 +1,23 @@
+import { Props } from '@storybook/addon-docs/blocks';
+import { Playground } from '../../../docs/Playground';
+import { Risk } from './Risk';
+
+# Risk
+
+This component will render a row of 7 rectangles with one colored rectangle containing. The colored rectangle contains a number from 1 - 7 depending on its position.
```
kmelkon, line 7: Grammar nazi: extra unnecessary word
```suggestion
This component will render a row of 7 rectangles with one colored rectangle. The colored rectangle contains a number from 1 - 7 depending on its position.
```

src/Molecules/Risk/Risk.tsx
```diff
@@ -0,0 +1,37 @@
...
+      ? p.theme.color.riskLevelActiveBackground
+      : p.theme.color.riskLevelInactiveBackground};
+`;
+
+export const Risk: React.FC<Props> = ({ risk = 0 }) => {
+  const clampedRisk = isNumber(risk) ? Math.max(1, Math.min(risk, 7)) : 0;
+
+  return (
+    <Flexbox container gutter={1}>
+      {R.range(1, 8).map(box => (
```
kmelkon, line 24: I suggest we take the range in as props to make this component more generic and updating clampedRaisk accordingly.

src/Molecules/Risk/Risk.tsx
```diff
@@ -0,0 +1,37 @@
...
+  justify-content: center;
+  align-items: center;
+  height: ${({ theme }) => theme.spacing.unit(4)}px;
+  background: ${p =>
+    p.isActive
+      ? p.theme.color.riskLevelActiveBackground
+      : p.theme.color.riskLevelInactiveBackground};
+`;
+
+export const Risk: React.FC<Props> = ({ risk = 0 }) => {
```
kmelkon, line 19: I suggest we name it something like a Scale unless @mjmonline has a better idea

src/Molecules/Risk/Risk.mdx
Vikingjonsson: thank you, will fix :)

src/Molecules/Risk/Risk.tsx
Vikingjonsson: new pr is up, thanks for reviewing

src/theme/theme.ts
```diff
@@ -234,6 +234,8 @@ export const createTheme = (config: ThemeConfig = {}): Theme => {
       pill8: rawColor.brandPink,
       pill9: rawColor.complementaryGreen1,
       pill10: rawColor.complementaryPink1,
+      riskLevelActiveBackground: rawColor.complementaryBlue1,
```
Vikingjonsson, line 237: will fix this names

src/Molecules/BarScale/BarScale.tsx
```diff
@@ -0,0 +1,119 @@
...
+  ${p => p.position === POSITIONS.FIRST && first}
+  ${p => p.position === POSITIONS.CENTER && center}
+  ${p => p.position === POSITIONS.LAST && last}
+  background: ${({ theme }) => theme.color.barScaleActiveBar};
+  color: ${({ theme }) => theme.color.textLight};
+`;
+
+const OFFSET_PERCENT = 50;
+const StyledFlexbox = styled(Flexbox).withConfig({
+  shouldForwardProp: prop => !['isActive'].includes(prop),
```
kmelkon, line 47: Looks fancy! need to read the new release notes to understand this :D

src/Molecules/BarScale/BarScale.types.ts
```diff
@@ -0,0 +1,12 @@
+export type Props = {
+  /** Needs to be between 0 and max  */
+  value: number;
+  /** The number of bars to render */
+  max: number;
```
mjmonline, line 5: is it not most likely to be 7? we can then have this as optional. If so don't forget to add `/** @default 7 */`

Vikingjonsson: atm it is most likely to be default seven, but in the future I it won't be

you think I should go ahead either way?

mjmonline: I think it's better to not make it required but rather optional

Vikingjonsson: okay i fix

src/Molecules/BarScale/BarScale.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+      <Flexbox container justifyContent="space-between">
+        <Flexbox item>Low</Flexbox>
+        <Flexbox item>Medium</Flexbox>
+        <Flexbox item>High</Flexbox>
+      </Flexbox>
+    }
+  />
+);
+
+export const BarScaleWithOverrideMaxLength = () => (
```
mjmonline, line 73: So because we use the storyshots as tests then you need to cover all your cases with stories. You can start by using the default story where you only have default values, in your case without the props `max` and `showValue`

src/Molecules/BarScale/BarScale.mdx
```diff
@@ -0,0 +1,45 @@
...
+<Props of={BarScale} />
+
+## Usage
+
+Basic usage
+
+```javascript
+<BarScale
+  value={3}
+  max={7}
```
mjmonline, line 21: not needed in basic usage

src/Molecules/BarScale/BarScale.tsx
```diff
@@ -0,0 +1,169 @@
...
+    p.isActive ? p.theme.color.barScaleActiveBar : p.theme.color.barScaleInactiveBar};
+
+  &::after {
+    display: ${p => (p.isActive ? 'block' : 'none')};
+    content: '';
+    position: absolute;
+    bottom: 100%;
+    margin-bottom: ${({ theme }) => theme.spacing.unit(TRIANGLE_OFFSET)}px;
+    left: ${OFFSET_PERCENT}%;
+    transform: translate(-${OFFSET_PERCENT}%);
```
mjmonline, line 77: the two rows above maybe doesn't have to be a const. The css only centers the triangle on the x-axis and the value is not used anywhere else

src/Molecules/BarScale/BarScale.tsx
```diff
@@ -0,0 +1,169 @@
...
+    width: 0;
+    height: 0;
+    border-left: ${TRIANGLE_SIDE_BORDER_SIZE}px solid transparent;
+    border-right: ${TRIANGLE_SIDE_BORDER_SIZE}px solid transparent;
+    border-top: ${({ theme }) => theme.spacing.unit(TRIANGLE_TOP_BORDER_SIZE)}px solid
+      ${({ theme }) => theme.color.barScaleActiveBar};
+  }
+`;
+
+const xAxisCollision = (a: HTMLElement, b: HTMLElement) =>
```
mjmonline, line 87: move types definitions to the types file. I know in mutations we just do it here but that was always the quick and dirty

src/Molecules/BarScale/BarScale.tsx
```diff
@@ -0,0 +1,169 @@
...
+  a.getBoundingClientRect().x <= b.getBoundingClientRect().x + b.getBoundingClientRect().width;
```
mjmonline, line 88: you could run `getBoundingClientRect()` just once and save the return to a const then use that const in the three places

src/Molecules/BarScale/BarScale.tsx
```diff
@@ -0,0 +1,169 @@
...
+  };
+
+  return (
+    <Box mt={10}>
+      <Flexbox container gutter={1}>
+        <IntersectionLeft ref={intersectionLeft} />
+        <IntersectionRight ref={intersectionRight} />
+
+        {R.range(1, verifiedMaxRating + 1).map(bar => (
+          <Flexbox key={bar} item flex="1">
```
mjmonline, line 133: does the `flex="1"` result the same value in all browsers?

src/Molecules/BarScale/BarScale.types.ts
```diff
@@ -0,0 +1,13 @@
+export type Props = {
+  /** Needs to be between 0 and max  */
+  value: number;
+  /** The number of bars to render */
+  /** @default 7 */
+  max: number;
+  /** show the number of the active bar */
+  showValue: boolean;
```
mjmonline, line 8: /** @default false */

src/Molecules/BarScale/BarScale.tsx
Vikingjonsson: true, but it is two different objects so I could only reuse it one

PR #506, opened 2020-04-15
Feat: update styled components

mjmonline:


yarn.lock
```diff
@@ -2223,6 +2223,13 @@
   dependencies:
     "@emotion/memoize" "0.7.3"
 
+"@emotion/is-prop-valid@^0.8.8":
```
nn-davsko, line 2226: Is it fine to have two versions?

KitsuneKelso, line 2226: I think I saw this brought up in another PR already, but was this resolution fine? Are you certain you don't want both versions to resolve to `0.8.8`? ☝️ 

I'll leave it up to your discretion. 🙂

PR #508, opened 2020-04-16
fix: remove collapsed warnings

robineng:


src/Molecules/CollapsibleCard/Collapsible.types.ts
```diff
@@ -6,5 +6,5 @@ export type CollapsibleProps = {
 };
 
 export type IndicatorsProps = {
-  collapsed: boolean;
+  $collapsed: boolean;
```
mjmonline, line 9: Is that dollar sign a hack to not pass the prop through to the actual html element? If so I think you should rather use `shouldForwardProp` that you can read more about here https://github.com/styled-components/styled-components/releases/tag/v5.1.0

JohnPhoto: Not really: 

> Think of transient props as a lightweight, but complementary API to shouldForwardProp. Because styled-components allows any kind of prop to be used for styling (a trait shared by most CSS-in-JS libraries, but not the third party library ecosystem in general), adding a filter for every **possible prop you might use can get cumbersome.**

mjmonline: Awesome, did not know this

PR #513, opened 2020-04-28
Feat/update card with title

gustav-pettersson-nordnet:
Update padding to <CardWithTitle> according to design

### DIFF
--- a/src/Molecules/CardWithTitle/CardWithTitle.tsx
+++ b/src/Molecules/CardWithTitle/CardWithTitle.tsx
@@ -1,17 +1,9 @@
 import React from 'react';
 import styled from 'styled-components';
-import { Card } from '../..';
+import { Card, Typography } from '../..';
 import { CardWithTitleComponent } from './CardWithTitle.types';
 import { Box } from '../../Atoms/Box/Box';
-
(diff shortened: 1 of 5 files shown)


src/Molecules/CardWithTitle/CardWithTitle.tsx
```diff
@@ -1,43 +1,35 @@
...
-   * `
-   * */
-  components: typeof components;
-} = props => {
+export const CardWithTitle: CardWithTitleComponent = props => {
   const { title, children, ...rest } = props;
 
   return (
     <StyledCard {...rest}>
-      <StyledHeader>{title}</StyledHeader>
```
mjmonline, line 35: `StyledHeader` used to render a `<header>` now it's a div. Intensional?

nn-davsko: Yeha I know, I am not sure it matters? It would be a semantic difference only as far as I can tell. Do you think I should bring it up with design?

smithsamyan: I didn't notice it but for future readers the box has as="header" now so it's not a div anymore.

PR #514, opened 2020-05-03
Add line scale

jesperwiner:


src/Molecules/LineScale/LineScale.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+  />
+);
+
+export const LineScaleAtRightEdge = () => (
+  <LineScale
+    value={100}
+    valueLabel="100.00"
+    averageValue={100}
+    averageValueLabel="Average for category 100.00"
+    minLabel="0"
```
nn-davsko, line 117: nit: I am not sure but since min and max -Labels are props perhaps have stories for when those are longer? Or perhaps it makes no sense, up to you.

src/Molecules/LineScale/LineScale.tsx
```diff
@@ -0,0 +1,223 @@
...
+    if (averageTextRef.current && intersectionLeft.current) {
+      const hasLeftCollision = xAxisCollision(averageTextRef.current, intersectionLeft.current);
+      setLeftAverageTextCollision(hasLeftCollision);
+    }
+
+    if (averageTextRef.current && intersectionRight.current) {
+      const hasRightCollision = xAxisCollision(intersectionRight.current, averageTextRef.current);
+      setRightAverageTextCollision(hasRightCollision);
+    }
+  }, [indicatorRef, averageTextRef, intersectionRight, intersectionLeft]);
```
smithsamyan, line 169: Question: Wouldn't adding value and valueLabel to this dependency array create support for dynamic redraws if the indicator value updates? Thinking specifically if the value updates and it's now in the left or right corner it will be cut into. 

I understand not redrawing on window resizing as that can be heavy but the value and valueLabels are both primitives so comparing them to their previous value should be possible

smithsamyan: UPDATE: I added the value and valueLabel and the indicator redrew correctly. Worth adding it? We could also add averageLabel and averageValue so a redraw happens then too.

src/Molecules/LineScale/LineScale.stories.tsx
jesperwiner: Can add that,np

jesperwiner:
> Looks good. I am not sure if voice over is a concern for this component but things aren't selectable with tab, then again I am not sure that's an issue with a component like this.

Didn't think about that, but I think the voice should only like say the value and average value, can look into that.

src/Molecules/LineScale/LineScale.tsx
jesperwiner: I tried that before, didn't get it to work I thought so I removed the dependencies again

src/Molecules/LineScale/LineScale.tsx
```diff
@@ -0,0 +1,229 @@
...
+          <Flexbox item flex="1 1 auto">
+            <StyledFlexbox
+              container
+              value={value}
+              averageValue={averageValue}
+              valueColor={valueColor}
+            >
+              <Indicator
+                tabIndex={-1}
+                aria-label={valueLabel}
```
mjmonline, line 186: aria-label is not needed here, the label is already assigned to be the content of this element. The result with be that the screenreader would read the label twice and maybe say something about a group?

src/Molecules/LineScale/LineScale.tsx
```diff
@@ -0,0 +1,229 @@
...
+          <IntersectionRight ref={intersectionRight} />
+          <Flexbox item flex="1 1 auto">
+            <StyledFlexbox
+              container
+              value={value}
+              averageValue={averageValue}
+              valueColor={valueColor}
+            >
+              <Indicator
+                tabIndex={-1}
```
mjmonline, line 185: Also not needed

src/Molecules/LineScale/LineScale.tsx
```diff
@@ -0,0 +1,229 @@
...
+            <Flexbox container justifyContent="space-between">
+              <Flexbox item>{minLabel}</Flexbox>
+              <Flexbox item>{maxLabel}</Flexbox>
+            </Flexbox>
+          )}
+          {(Boolean(averageValue) || Boolean(averageValueLabel)) && (
+            <Flexbox item flex="1 1 auto">
+              <Flexbox container>
+                <AverageLabel
+                  tabIndex={0}
```
mjmonline, line 209: Because you can't really interact with this element then it shouldn't be tabable, so no need to set a tabindex

src/Molecules/LineScale/LineScale.tsx
```diff
@@ -0,0 +1,229 @@
...
+                  aria-label={averageValueLabel}
```
mjmonline, line 210: also not needed

src/Molecules/LineScale/LineScale.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+  parameters: {
+    ...docs.parameters,
+  },
+  decorators: [withKnobs],
+};
+
+const getLineScaleProps = ({
+  value = 50,
+  valueLabel = '50',
+  valueColor = '#FF0000',
```
mjmonline, line 17: Is is some default behavior that if it's 50% then we have a different color on the indicator? If se then maybe have a prop inside the component with a default value of like negative from the theme

jesperwiner: Its just a story to show a different color, nothing to do with the value 50

PR #516, opened 2020-05-06
fix: add co2 icon

jesperwiner:


iamstarkov:
:shipit:

jesperwiner:
> Would have liked a name that's not associated with what it's used for but rather for what the icon is. In case we have a different way of presenting co2 later but you decide if you want to change it

Will change it to "leaf" I agree its better

PR #517, opened 2020-05-06
feat: implement slider component

crewneck:


src/Molecules/Slider/Slider.types.ts
```diff
@@ -0,0 +1,18 @@
...
+type ColorFn = (t: Theme) => Values<Theme['color']>;
+
+export type Props = {
+  onChange: (v: number) => void;
+  min: number;
+  max: number;
+  step: number;
+  value: number;
+  leftColor?: ColorFn;
+  rightColor?: ColorFn;
```
KitsuneKelso, line 17: Just an idea, but maybe these should be `activeColor` and `inactiveColor` or something like that instead.

src/Molecules/Slider/Slider.tsx
```diff
@@ -0,0 +1,208 @@
...
+    }
+    if (newX >= min && newX <= max) {
+      if (typeof onChange === 'function') {
+        onChange(newX);
+      }
+    }
+  };
+
+  const gradient = {
+    background: `linear-gradient(to right, ${
```
KitsuneKelso, line 178: I did a quick test in Edge now and noticed that the `linear-gradient` doesn't seem to want to play along there. The interaction itself is working as intended, but it's not getting the background color at all.
Screenshot 2020-05-06 at 15 49 25

src/Molecules/Slider/Slider.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+import Slider from '.';
+import { Typography, Number } from '../..';
+
+export default {
+  title: 'Molecules | Slider',
+  decorators: [withKnobs],
+};
+
+const getSliderProps = ({ min = 0, max = 100, step = 1 }) => ({
+  min: number('Minumum', min),
```
anneback, line 12: typo, should be Minimum.

PR #526, opened 2020-05-08
feat: 🎸 add colorSet prop to Development

robineng:


mjmonline:
The way we handle colors with this PR is unique in the library. The question would then be should this behavior be used in other components? Feels like we are over engineering how to solve a hack of when designers don't want to use the high contrast colors. I would rather like to expose something like `positiveColor` and `negativeColor` props and override them that way, what do you say @robineng?

PR #527, opened 2020-05-11
fix: 🐛 slider - prevent text selection during thumb button drag

crewneck:


src/Molecules/Slider/Slider.tsx
```diff
@@ -109,6 +109,9 @@ const Slider: FC<Props> = ({ onChange, value, max, min, step, leftColor, rightCo
   };
 
   const handleMouseMove = (event: MouseEvent) => {
+    event.stopPropagation();
+    event.preventDefault();
```
KitsuneKelso, line 113: This might need to be added to the `handleTouchMove` as well.

PR #534, opened 2020-05-14
Update beta branch

gustav-pettersson-nordnet:


gustav-pettersson-nordnet:
> What was the original cause for this merge? Eventually it won't matter but it might be good to keep the beta-branch only focused on the migration of the CardWithTitle and CardWithTabs features. Is there something in master that is needed here?

There was a few. One exampke was: @nordnet/ui/dist/types"' has no exported member 'Slider'.

PR #550, opened 2020-05-28
feat: 🎸 implement ProgressBar

KitsuneKelso:
Screenshot 2020-05-28 at 13 50 55


src/Molecules/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const stepBubble = (stepNumber: number) => {
+    const stepDone = stepNumber < currentStep;
+    const stepActive = stepNumber === currentStep;
+
+    const titleProgress = () => {
+      if (stepDone) {
+        return 'step done';
+      }
+      if (stepActive) {
+        return 'active step';
```
mjmonline, line 61: proper way of handling this is to use `aria-active="step"` https://www.aditus.io/aria/aria-current/

src/Molecules/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,118 @@
...
+      return 'step not done';
+    };
+
+    const title = `${
+      stepLabels ? stepLabels[stepNumber - 1] : stepNumber.toString()
+    } - ${titleProgress()}`;
+
+    return (
+      <React.Fragment key={stepNumber}>
+        <StyledBubble
```
mjmonline, line 72: so because these steps is in a pre defined order we should use `<ol>` with `<li>` wrapping each step

src/Molecules/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,118 @@
...
+          container
+          item
+          shrink={0}
+          alignItems="center"
+          justifyContent="center"
+          done={stepDone}
+          active={stepActive}
+          title={title}
```
mjmonline, line 80: would prefer to use `VisuallyHidden` from UI to consistently let the screen reader for example say "step x is not done"

src/Molecules/ProgressBar/ProgressBar.tsx
mjmonline: You also need to wrap your `<ol>` in a `<div role="group" title="progress bar">`

src/Molecules/ProgressBar/ProgressBar.types.ts
```diff
@@ -0,0 +1,8 @@
+export type Props = {
```
mjmonline, line 1: for some reason your props is not being rendered on the docs page

src/Molecules/ProgressBar/ProgressBar.tsx
mjmonline: You also need to wrap your <ol> in a <div role="group" title="progress bar">

mjmonline: You also need to wrap your `<ol>` in a `<div role="group" title="progress bar">`

src/Molecules/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,118 @@
...
+        >
+          <Typography type="secondary" color={t => t.color.textLight} aria-hidden>
+            {stepDone ? <Icon.CheckMark color={t => t.color.textLight} size={4} /> : stepNumber}
+          </Typography>
+          {showStepLabels && (
+            <StyledTypography
+              type="secondary"
+              weight={stepActive ? 'bold' : 'regular'}
+              color={t => t.color.text}
+              aria-hidden
```
mjmonline, line 90: Why aria-hidden here?

src/Atoms/Flexbox/Flexbox.types.ts
```diff
@@ -45,6 +45,7 @@ export type ContainerProps = {
 
   /** a11y */
   role?: string;
+  title?: string;
```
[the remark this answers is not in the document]
KitsuneKelso: This is to get support for both the title you get when you hover over the elements as well as to get screen reader support when tabbing between the steps.

Screenshot 2020-05-28 at 16 24 14

Excuse the horrible screencap, it was hard to get one with a title visible. 😅

src/Molecules/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,118 @@
...
+  shouldForwardProp: prop => !['done', 'active'].includes(prop),
+})<{ done: boolean; active: boolean }>`
+  position: relative;
+  border-radius: 50%;
+  width: ${({ theme }) => theme.spacing.unit(8)}px;
+  height: ${({ theme }) => theme.spacing.unit(8)}px;
+  color: ${({ theme }) => theme.color.progressBarText};
+  background: ${({ theme, done, active }) => {
+    if (done) {
+      return theme.color.progressBarDone;
```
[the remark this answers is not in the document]
KitsuneKelso: Will add this!

src/Molecules/ProgressBar/ProgressBar.types.ts
KitsuneKelso: Strange, I'll look into it, I'm guessing there's some small detail missing.

src/Molecules/ProgressBar/ProgressBar.tsx
KitsuneKelso: The only reason is essentially that we wanted only the main "bubble" to be the focusable element for the screen reader, when the children weren't hidden it would navigate to the whole bubble and read the title, then jump to the number in the bubble, then down to the label.

src/Molecules/ProgressBar/ProgressBar.tsx
KitsuneKelso: We were debating setting it up as a list as well, the only problem is the way we chose to solve the line between the "items", how would we keep that out of the list of items:
Screenshot 2020-05-28 at 16 31 39

I'll experiment a bit with how the line is done and see if I can convert the steps into list items instead. 👍

mjmonline: Try a `::after` on `<li>`

src/Atoms/Flexbox/Flexbox.types.ts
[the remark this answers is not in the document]
mjmonline: Ok then maybe we should just pass `...rest` because the title should/could be available on all components. That way we don't have to define similar props like that on all components

src/Molecules/ProgressBar/ProgressBar.types.ts
KitsuneKelso: Got them working now, and expanded the docs while I was at it. 🙂

src/Molecules/ProgressBar/ProgressBar.tsx
mjmonline: But this will essentially make the label not accessible by the screen reader. It's like putting a `display: none` for regular users

KitsuneKelso: Exactly, that's what I want, I only want the screen reader to tab through the full progress bar steps. That's why the `title` on the container should contain the full interpretable text to reflect each step. Otherwise it will tab through first the container, then the step number, then the label.

src/Molecules/ProgressBar/ProgressBar.tsx
KitsuneKelso: @mjmonline Relating to the `aria-hidden` props on the children of this element, this is the only part that I want the screen reader to see. ☝️

KitsuneKelso:
@mjmonline do you think this is alright as an MVP 1 release, or would you like any additional changes before we proceed?

src/Atoms/Flexbox/Flexbox.types.ts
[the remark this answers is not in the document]
mjmonline: Any comment on this?

src/Molecules/ProgressBar/ProgressBar.tsx
mjmonline: Any comment on this?

mjmonline: @KitsuneKelso Any comment on this?

src/Molecules/ProgressBar/ProgressBar.tsx
mjmonline: @KitsuneKelso Any comment on this?

src/Atoms/Flexbox/Flexbox.types.ts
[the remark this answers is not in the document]
mjmonline: @KitsuneKelso Any comment on this?

src/Molecules/ProgressBar/ProgressBar.tsx
KitsuneKelso: I'm not sure how this one is supposed to work, is only the actual active step supposed to get that aria prop?

src/Atoms/Flexbox/Flexbox.types.ts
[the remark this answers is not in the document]
KitsuneKelso: I tried spreading `...rest` in the Flexbox component instead but I only kept running into type violations, even when I did it the exact same way as other components and types have handled it, so I'm not sure how to solve it.

KitsuneKelso:
> There is still some questions above, please check them :) Also will there be a scenario where the steps is clickable?

Yes! That will be a feature added later, see ticket DND-469 in the backlog.

src/Molecules/ProgressBar/ProgressBar.tsx
KitsuneKelso: @mjmonline I've been attempting to convert the elements to `<ol>` and `<li>` for a while now, but I'm running into some serious problems with how `styled-components` generates the `:after` pseudo-elements, they get generated on the same level as the `<span>` children of the `<li>` instead as a sibling of the `<li>` itself:
Screenshot 2020-06-03 at 14 33 07
This is making it practically impossible to style it and I'm now sitting with a jumbled mess of spaghetti CSS trying to style the elements. 😅

It's honestly starting to seem like a lot more work than it's worth to have these as list elements, can that be a potential improvement for the future instead, or do you see it as a blocker?

KitsuneKelso: I've got it working now semi-well:
Screenshot 2020-06-03 at 14 45 12

The caveat now is that each `<li>` element needs the following styling to get it to work this way:
```js
const StyledBubble = styled.li.withConfig({
  shouldForwardProp: prop =>
    ![
      'done',
      'active',
      'colorDone',
      'colorActive',
      'colorNext',
      'stepNumber',
      'numberOfSteps',
    ].includes(prop),
})<{
  done: boolean;
  active: boolean;
  colorDone: Props['colorDone'];
  colorActive: Props['colorActive'];
  colorNext: Props['colorNext'];
  stepNumber: number;
  numberOfSteps: number;
}>`
  list-style-type: none;
  & > span:first-of-type {
    display: flex;
    align-items: center;
    justify-content: center;
    flex-shrink: 0;
    position: relative;
    border-radius: 50%;
    width: ${({ theme }) => theme.spacing.unit(8)}px;
    height: ${({ theme }) => theme.spacing.unit(8)}px;
    color: ${({ theme }) => theme.color.progressBarText};
    background: ${({ theme, done, active, colorDone, colorActive, colorNext }) => {
      if (done) {
        return colorDone ? colorDone(theme) : theme.color.progressBarDone;
      }
      if (active) {
        return colorActive ? colorActive(theme) : theme.color.progressBarActive;
      }
      return colorNext ? colorNext(theme) : theme.color.progressBarNext;
    }};
  }
  ::after {
    content: "";
    display: ${({ stepNumber, numberOfSteps }) => (stepNumber < numberOfSteps ? 'block' : 'none')};
    position: absolute;
    transform: translate(${({ theme }) => theme.spacing.unit(8)}px, -${({ theme }) =>
  theme.spacing.unit(4)}px);
    width: calc((100% / ${({ numberOfSteps }) => numberOfSteps - 1}) - ${({ theme }) =>
  theme.spacing.unit(8)}px);
    height: 2px;
    background: ${({ theme, done, colorDone, colorNext }) => {
      if (done) {
        return colorDone ? colorDone(theme) : theme.color.progressBarDone;
      }
      return colorNext ? colorNext(theme) : theme.color.progressBarNext;
    }}
`;
```

That's quite nasty if you ask me. 😬

crewneck:
> There is still some questions above, please check them :) Also will there be a scenario where the steps is clickable?

@mjmonline Can you be more specific on what your missing from an mvp1?

src/Molecules/ProgressBar/ProgressBar.tsx
mjmonline: Well in the UI semantic html and accessibility is important. For the user a few extra lines of css would not make any difference but on the other hand having it as a list `<ol>` would give the user all the benefits that a list gives

PR #551, opened 2020-05-28
Feat: add accordion component

mjmonline:


src/Molecules/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const Footer = styled.div`
+  padding-top: ${p => p.theme.spacing.unit(3)}px;
+`;
+
+export const Accordion: Component = ({ children, footer }) => {
+  return (
+    <>
+      {children}
+      {footer ? (
```
crewneck, line 14: Can you do it without a ternary here. Use && instead?

PR #552, opened 2020-05-29
feat: 🎸 add bolean for closing modal on backdrop click

crewneck:


src/Molecules/Modal/Modal.types.ts
```diff
@@ -5,6 +5,7 @@ export type InnerProps = {
   title?: React.ReactNode;
   footer?: React.ReactNode;
   hideClose?: boolean;
+  closeOnBackdropClick?: boolean;
```
mjmonline, line 8: add jsdoc saying it's false by default like `/** @default false */`

src/Molecules/Modal/ModalInner.tsx
```diff
@@ -111,6 +113,8 @@ export const ModalInner: React.FC<Props> = ({
   const seed = useUIDSeed();
   const titleId = seed('ModalTitle');
   const hasHeader = !hideClose || title;
+  const ref = useRef<HTMLDivElement>(null);
+  useOnClickOutside(ref, onClose as () => void);
```
mjmonline, line 117: Feel like it would be more performant to just add a click event on the backdrop itself rather than to listen to clicks on all elements except the Modal

crewneck: That doesn't work since clicking inside the dialog area will also close the modal. Feels safer and and less intrusive to do this to avoid start changing in the z-indexes for the different parts of the component.

PR #553, opened 2020-05-29
Feat/table row

nn-davsko:
Table Row first iteration


src/Molecules/Table/Row/Row.types.ts
```diff
@@ -0,0 +1,15 @@
+import React, { ReactNode } from 'react';
+
+type HtmlProps = {} & React.HTMLAttributes<HTMLDivElement>;
+
+export type Size = 's' | 'm' | 'l';
+
+type Props = {
+  size?: Size;
```
mjmonline, line 8: add jsdoc for default value

mjmonline: same for the rest of the props with a default value

src/Molecules/Table/Table.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+  </Table>
+);
+
+export const TableExpanded = () => {
+  const TableExpandedExample = () => {
+    const [expanded, setExpanded] = useState(false);
+    return (
+      <Table>
+        <Table.Row>Default</Table.Row>
+        <Table.Row expanded={expanded} expandableContent={<div>I am expanded</div>}>
```
mjmonline, line 28: Would be nice to have a uncontrolled version of the Row component where it handles itself. That requires though that the toggle button is baked in and maybe replaceable with a prop. I think it's a good idea to have the toggle baked in so that we don't have to create it (and create it correctly) every time

PR #565, opened 2020-06-09
Feat/table header

smithsamyan:
Added a table header component


src/Atoms/Icon/__snapshots__/Icon.stories.storyshot
```diff
@@ -2963,6 +2963,49 @@ Array [
...
+  fill: #0046FF;
+  -webkit-flex-shrink: 0;
+  -ms-flex-negative: 0;
+  flex-shrink: 0;
+  display: block;
+}
+
+.c1 {
+  -webkit-transform: rotate(undefineddeg);
+  -ms-transform: rotate(undefineddeg);
```
mjmonline, line 2982: something fishy here

src/Atoms/Icon/components/SortArrow.tsx
```diff
@@ -0,0 +1,31 @@
...
+  direction: 'ascending' | 'descending';
+};
+
+const DIRECTION = {
+  ascending: 180,
+  descending: 0,
+};
+
+const StyledIconBase = styled(IconBase)<Direction>`
+  transform: ${p => `rotate(${DIRECTION[p.direction]}deg`});
```
mjmonline, line 16: `${DIRECTION[p.direction]}` rendering undefined

src/Molecules/Table/shared/constants.ts
```diff
@@ -0,0 +1,9 @@
+import { theme } from '../../..';
+
+export const DENSITY_PADDING_SMALL = theme.spacing.unit(0);
```
mjmonline, line 3: unnecessary function call? 0 should be enough?

src/Molecules/Table/Header/Header.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (cellFlexProps) {
+      columnDispatch({ type: ACTION_SET_FLEX_PROPS, flexProps: cellFlexProps });
+    }
+  }, [cellFlexProps, columnDispatch]);
+
+  return (
+    <Flexbox
+      className={className}
+      role="columnheader"
+      {...ariaSorted}
```
mjmonline, line 85: add `scope="col"` to suggest header for a column

src/Molecules/Table/Header/HeaderContent/HeaderContent.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return (
+      <TextWrapper fontSize={fontSize} density={density} sorted={false}>
+        {children}
+      </TextWrapper>
+    );
+  }
+
+  return (
+    <StyledLink
+      href="#"
```
mjmonline, line 33: You do not want a link here, should be a `<button>` because you're not navigating the user anywhere. I get that it has to do with styling but accessibility and user experience is more important

src/Molecules/Table/Row/Row.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import styled from 'styled-components';
+import Flexbox from '../../../Atoms/Flexbox';
+import { RowComponent } from './Row.types';
+import { Box } from '../../../index';
```
mjmonline, line 5: Not important but `Box` and `Flexbox` could be imported from same path

src/Molecules/Table/Header/HeaderContent/HeaderContent.tsx
```diff
@@ -0,0 +1,50 @@
...
+  if (!sortable || sortOrder === null) {
+    return (
+      <TextWrapper fontSize={fontSize} density={density} sorted={false}>
+        {children}
+      </TextWrapper>
+    );
+  }
+
+  return (
+    <StyledLink
```
nn-davsko, line 32: This needs to be a button for accessibility reasons, probably need to do some changes to our current button to make it work.

src/Molecules/Table/Row/Row.tsx
```diff
@@ -0,0 +1,56 @@
...
+      role="row"
+      expanded={expanded}
+      separatorColor={separatorColor}
+      {...htmlProps}
+    >
+      <Flexbox container alignItems="center">
+        {children}
+      </Flexbox>
+      {expanded && (
+        <ExpandedContent>
```
mjmonline, line 48: no styling and not doing anything else with that component, maybe just use a `div`?

src/Molecules/Table/RowGroup/RowGroup.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { Props } from './RowGroup.types';
+import { Flexbox } from '../../../index';
+
+export const RowGroup: React.FC<Props> = ({ children }) => {
+  return <Flexbox role="rowgroup">{children}</Flexbox>;
```
mjmonline, line 6: needs `item` or `container`? I think otherwise this is just a div

src/Molecules/Table/Table.types.ts
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { RowComponent } from './Row/Row.types';
+import { HeaderComponent } from './Header/Header.types';
+import { RowGroupComponent } from './RowGroup/RowGroup.types';
+import { constants } from './shared';
+
+type HtmlProps = {} & React.HTMLAttributes<HTMLDivElement>;
+
+export type Props = {
+  className?: string;
```
mjmonline, line 10: Do we want to be able to give the table a title? Maybe `<caption>{title}</caption>` inside the `<table>` would read out a title for the table. Having a heading before the table does not really connect it to the table unless we use `aria-labeledby="id-of-the-heading-before-the-table"`. It's not a showstopper, we could still go without the clear connection, just something to think about

src/Atoms/Icon/components/SortArrow.tsx
gustav-pettersson-nordnet: handle case when DIRECTION[p.direction] is undefined

src/Molecules/Table/RowGroup/RowGroup.tsx
smithsamyan: Good point, we need to remove this anyway, this file was added for a11y but it doesn't seem to do anything and a lot of accessible tables don't include it so we should remove it

src/Molecules/Table/Table.types.ts
smithsamyan: Yes. There's a task to do that, I think we should go the caption route

src/Molecules/Table/Header/Header.tsx
gustav-pettersson-nordnet: Do we need both role="columnheader" and scope="col"? Do they have different purposes or do they both "suggest header for a column"?

mjmonline: Oh my bad, scope col would apply to a `th`. In this case it works fine!

PR #566, opened 2020-06-10
Add big card version

Vikingjonsson:
Added variant to props for cards to use extra padding and a larger heading on CardWithTabs and CardWithTitles.
The design document calls this version Big.

### DIFF
--- a/src/Molecules/CardWithTitle/CardWithTitle.stories.tsx
+++ b/src/Molecules/CardWithTitle/CardWithTitle.stories.tsx
@@ -65,14 +65,14 @@ withComponentAsTitleStory.story = {
   name: 'With Component As Title',
 };
 
-export const withExtraPadding = () => (
+export const withVariantBig = () => (
   <CardWithTitle variant="big" title="Title for the card">
     A CardWithTitle with extra padding
   </CardWithTitle>
 );
 
-withExtraPadding.story = {
-  name: 'With Extra Padding',
+withVariantBig.story = {
+  name: 'With variant big',
 };
 
 export const cardWithTitleAsArticleMostCardWithTitleShouldBeArticle = () => (
--- a/src/Molecules/CardWithTitle/CardWithTitle.stories.tsx
+++ b/src/Molecules/CardWithTitle/CardWithTitle.stories.tsx
@@ -65,6 +65,16 @@ withComponentAsTitleStory.story = {
   name: 'With Component As Title',
 };
 
+export const withExtraPadding = () => (
+  <CardWithTitle padded title="Title for the card">
+    A CardWithTitle with extra padding
+  </CardWithTitle>
+);
+
+withExtraPadding.story = {
+  name: 'With Extra Padding',
+};
+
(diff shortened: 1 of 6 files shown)


src/Organisms/CardWithTabs/CardWithTabs.stories.tsx
```diff
@@ -50,6 +50,21 @@ withInitialActiveTabIndex.story = {
   name: 'with initialActiveTabIndex',
 };
 
+export const withVariantBig = () => (
+  <CardWithTabs variant="big" title="Title for the card" initialActiveTabIndex={0}>
```
mjmonline, line 54: We need to give the developer information that this component takes a variant prop, right now this works because `CardWithTabs` spreads `...rest` to `CardWithTitle`. A developer would not know that unless they read the source code. So we need to add this as a prop in types so it can be shown in storybook docs and as suggestion in the code editors

src/Molecules/CardWithTitle/CardWithTitle.stories.tsx
```diff
@@ -65,6 +65,16 @@ withComponentAsTitleStory.story = {
...
 };
 
+export const withExtraPadding = () => (
+  <CardWithTitle variant="big" title="Title for the card">
+    A CardWithTitle with extra padding
+  </CardWithTitle>
+);
+
+withExtraPadding.story = {
+  name: 'With Extra Padding',
```
mjmonline, line 75: This prop changes font size as well so maybe just name it "With variant big" or something

src/Organisms/CardWithTabs/CardWithTabs.stories.tsx
```diff
@@ -50,6 +50,21 @@ withInitialActiveTabIndex.story = {
...
+      1
+    </CardWithTabs.Tab>
+    <CardWithTabs.Tab title="Tab title 2" onTitleClick={action('Clicked title2')}>
+      2
+    </CardWithTabs.Tab>
+  </CardWithTabs>
+);
+
+withVariantBig.story = {
+  name: 'with Variant Big',
```
mjmonline, line 65: same here with the name

src/Organisms/CardWithTabs/CardWithTabs.stories.tsx
Vikingjonsson: I'll fix that, thanks

src/Molecules/CardWithTitle/CardWithTitle.stories.tsx
Vikingjonsson: Ops missed to change that one, thanks! will fix it

src/Organisms/CardWithTabs/CardWithTabs.stories.tsx
Vikingjonsson: I do get suggestions about variant and type now in the tabbed version. Will add it so it shows up in docs

mjmonline: My bad, `CardWithTabs` already imports all props from `CardWithTitle`

PR #567, opened 2020-06-10
Disco 2584

wiktoriatomzik:


src/Atoms/Icon/index.ts
```diff
@@ -119,16 +126,21 @@ export default {
   Financial,
   Flag,
   Globe,
+  Guided,
```
Vikingjonsson, line 129: is it possible to give this a more descriptive name?

PR #569, opened 2020-06-11
Feat/table cell

gustav-pettersson-nordnet:


src/Molecules/Table/Cell/Cell.tsx
```diff
@@ -0,0 +1,35 @@
...
+const Cell: CellComponent = ({ children, className, density, fontSize, columnId }) => {
+  const [columnState] = useColumn(columnId);
+
+  if (!R.prop('flexProps', columnState)) {
+    return null;
+  }
+
+  return (
+    <StyledFlexbox className={className} role="cell" {...columnState.flexProps}>
+      {isElement(children) && children}
```
nn-davsko, line 23: need a story for this

src/Molecules/Table/Header/HeaderContent/TextWrapper.tsx
```diff
@@ -1,42 +1,27 @@
...
 })<{ density: Density }>`
   padding-top: ${p => getDensityPaddings(p.density)}px;
   padding-bottom: ${p => getDensityPaddings(p.density)}px;
   display: inline-block;
+  word-break: break-word;
 `;
 
-export const TextWrapper: React.FC<TextWrapperProps> = ({
-  fontSize,
+export const TextWrapper: TextWrapperComponent = ({
```
jesperwiner, line 17: Im thinking if it might be confusing later on to have multiple wrapper components named the same.

smithsamyan: Maybe but there are different TextWrappers for different components. They're accessed differently too. In this case it would be 
`<Table.Header.TextWrapper>`

PR #573, opened 2020-06-15
fix: revert search drawer alignment

Vikingjonsson:


src/Molecules/Drawer/Drawer.tsx
```diff
@@ -8,6 +8,7 @@ import { Props, TitleProps } from './Drawer.types';
 import { isBoolean, isElement } from '../../common/utils';
 import { Typography, Icon, useKeyPress, Portal, useMedia, Button } from '../..';
 
+const CROSS_SIZE = 5;
```
kmelkon, line 11: Why do you need two constants for the same number?

mjmonline: They don't do the same things, it just happens that both is 5

PR #580, opened 2020-06-18
Feat/hiding columns

oscarsandberg:
MVP1: Simple implementation to hide columns in a table. Naive implementation through CSS (display: none), therefore columns still exists in DOM. Should be revisited if performance becomes an issue.

Wait to merge until other branches are merged.


smithsamyan:
This has been superseded by feat/table-hide-column due to it including more code than needed including performance changes that has already been merged.

PR #584, opened 2020-06-18
fix/performance-table

smithsamyan:
Increases table performance by 

* Splitting context into data and layout. This prevents unnecessary renders due to data changes rather than layout changes. 
* Memoizing Cell to skip re-render faster after sorting
* Update big table story to show how it should be written in a performant way
* Add an example of how to virtualize table using react-window


smithsamyan:
Oops, wrong one to close

PR #586, opened 2020-06-22
Feat/table sticky header

gustav-pettersson-nordnet:
Sticky header. Works on latest browsers. Degrades gracefully on IE11 without sticky.


src/Molecules/FlexTable/shared/FlexTableProvider/FlexTableProvider.types.ts
```diff
@@ -2,10 +2,12 @@ import { Density } from '../shared.types';
 
 export type FlexTableState = {
   density: Density;
+  stickyHeader: boolean;
 };
 
 export type Props = {
```
smithsamyan, line 8: Maybe set Props as {} & FlexTableState ?

PR #588, opened 2020-06-22
feat: 🎸 add controlled expand, rename prop to add expand on row

smithsamyan:
Add expand that can be controlled from the outside. Also change name of prop that adds an expand cell


fredriklu:
Discovered some bugs, will investigate and fix

PR #589, opened 2020-06-22
Feat/table hide column

oscarsandberg:
MVP1: Simple implementation to hide columns in a table. Naive implementation through CSS (display: none), therefore columns still exists in DOM. Should be revisited if performance becomes an issue.


gustav-pettersson-nordnet:
Closing and reopening to trigger travis to use node version 12 specified in the .travis.yml file

src/Atoms/Flexbox/Flexbox.types.ts
```diff
@@ -8,6 +8,8 @@ export type ContainerProps = {
   direction?: 'row' | 'row-reverse' | 'column' | 'column-reverse';
   wrap?: 'nowrap' | 'wrap' | 'wrap-reverse';
   gutter?: number;
+  hidden?: boolean;
+  title?: string;
```
smithsamyan, line 12: Why have we added title here?

mjmonline: Agree

PR #594, opened 2020-06-23
Fix/table expanded value

gustav-pettersson-nordnet:
For expanded content: 12px for both label and value for compact view (smallest density).


src/Molecules/FlexTable/FlexTable.stories.tsx
```diff
@@ -686,6 +677,52 @@ const expandedItemsGenerator = (renderComponent = false) =>
...
+          <FlexTable.Cell columnId="column2">Expandable</FlexTable.Cell>
+          <FlexTable.Cell columnId="column3">Expandable</FlexTable.Cell>
+        </FlexTable.Row>
+      </FlexTable>
+    );
+  };
+  return <ExpandedTableExample />;
+};
+
+export const TableWithDifferentRows = () => (
```
smithsamyan, line 717: Im not sure we should keep this one. There shouldn't be a table without cells.

src/Molecules/FlexTable/Row/ExpandItems/TextWrappers.tsx
```diff
@@ -16,13 +17,18 @@ export const TextWrapperLabel: React.FC = ({ children }) => (
...
-);
+export const TextWrapperValue: React.FC = ({ children }) => {
+  const { density } = useFlexTable();
+  return (
+    <>
+      <Media query={t => t.media.lessThan(t.breakpoints.md)}>
+        <StyledTypography type="tertiary">{children}</StyledTypography>
+      </Media>
+      <Media query={t => t.media.greaterThan(t.breakpoints.md)}>
+        <StyledTypography type={density === 's' ? 'tertiary' : 'secondary'}>
```
smithsamyan, line 28: Fontsize should be decided separately from density. This should instead look into the font size rather than using the density setting

src/Molecules/FlexTable/FlexTable.stories.tsx
nn-davsko: I agree, also the name of the story is a bit broad, if we want to show all different configurations for a row then perhaps its suitable but I don't think the story covers that at the moment

fredriklu:
This PR has now been updated to also include:

- `Expandable` to context
- Expandable padding
- `FontSize` Change
- `HeaderSeparator` color recreated

PR #596, opened 2020-06-24
Feat/truncate with tooltip

smithsamyan:
Enable truncation of content in cell or header with a tooltip


src/Molecules/FlexTable/Cell/Cell.tsx
```diff
@@ -25,7 +26,11 @@ const Cell: CellComponent = React.memo(({ children, className, fontSize, columnI
       {isElement(children) && children}
       {isFunction(children)
         ? children({ fontSize, columnId })
-        : !isElement(children) && <TextWrapper fontSize={fontSize}>{children}</TextWrapper>}
+        : !isElement(children) && (
+            <TruncateWithTooltip label={children} position="top">
+              <TextWrapper fontSize={fontSize}>{children}</TextWrapper>
```
nn-davsko, line 31: I like it but we have to consider what we export to the user if they are to make a custom element. For instance if they want a custom header or cell they now need one more building block to get the "plain vanilla" function before adding their own. This has nothing to do with the code above just a thought about how we expose building blocks if we start to get a lot of them.

src/Molecules/FlexTable/Cell/TextWrapper.tsx
```diff
@@ -1,18 +1,10 @@
...
-const StyledTypography = styled(Typography)`
-  display: inline-block;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-  overflow: hidden;
-`;
-
 export const TextWrapper: TextWrapperComponent = ({ fontSize = 'm', children }) => (
-  <StyledTypography type={getFontSizeTypographyType(fontSize)} color={t => t.color.text}>
+  <Typography type={getFontSizeTypographyType(fontSize)} color={t => t.color.text}>
```
nn-davsko, line 7: Nice!

src/Molecules/FlexTable/Cell/Cell.tsx
smithsamyan: I have moved it now to text wrapper instead as that's the most common area it should be used anyway

PR #597, opened 2020-06-25
Fix title breaking to several lines

nn-davsko:
Also removed the box i don't like the extra div compared to a small styled component, I know we are not consistent in this...


smithsamyan:
Truncate tooltip PR will remove this feature as it adds truncation to headers :)

smithsamyan:
Oh my bad!

PR #598, opened 2020-06-25
Feat/truncation components

smithsamyan:
Add two components to handle truncation. One for truncating text and one that truncates text and adds a tooltip when text is truncated


smithsamyan:
You can set maxWidth with styled components, thats why I choose to not include it

PR #601, opened 2020-06-26
Fix/table typescript errors

fredriklu:
Recast `CellComponent`, `sortOrder` undefined check, and changed error message for `ColumnProvider`


src/Molecules/FlexTable/Cell/Cell.tsx
```diff
@@ -30,7 +28,8 @@ const Cell: CellComponent = React.memo(({ children, className, columnId }) => {
         : !isElement(children) && <TextWrapper fontSize={fontSize}>{children}</TextWrapper>}
     </StyledFlexbox>
   );
-});
+};
 
 Cell.TextWrapper = React.memo(TextWrapper);
-export default Cell;
+// React.memo/NamedExoticComponent prevents TextWrapper from being included in types, recast to retain type.
+export default (React.memo(Cell) as any) as CellComponent;
```
smithsamyan, line 35: Maybe instead of as Any we can export it as ReactNode or something? otherwise looks good

src/Molecules/FlexTable/Header/Header.tsx
```diff
@@ -64,8 +64,7 @@ const Header: HeaderComponent = props => {
 
   useEffect(() => {
     // If the sortOrder changes from the outside, update internal the column sort state
-    if (controlledSort) {
-      // @ts-ignore
+    if (controlledSort && sortOrderProp !== undefined) {
```
smithsamyan, line 67: This solution works, however controlledSort currently checks that sortOrderProps isn't undefined. Feels weird that we double check pretty much the same thing just for typescript. I don't have any other solutions though so maybe add a comment or set the type some other way?

fredriklu: We tried a lot of other things that didn't quite work, e.g. type guard that didn't work. We think that TS isn't smart enough to realise the that connection since we're setting sortOrder: sortOrderProp later. Of course a comment can be added though!

robinandersson: Indeed, as @fredriklu mentions, the issue is that typeguard results are not stored in variables in TS (yet). It's explained in more detail here, with some possible workarounds if we want to it more correctly:

[Typescript Check for undefined and assign to variable](https://stackoverflow.com/a/56907826)

PR #605, opened 2020-06-29
feat(Timeline): adds Timeline component

anneback:
I gave it a try on my own with completing the Timeline with theory crafting from @KitsuneKelso. We still have a11y to take care of. But I need to component's functionality to fill it with data.


src/Molecules/Timeline/Timeline.stories.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { Box } from '../..';
+import Timeline from './Timeline';
+// import { Statuses } from './Timeline.types';
```
edenvin, line 4: Hello, comment!

src/Molecules/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,134 @@
...
+`;
+
+const Timeline: React.FC<Props> = ({ steps, colorSuccess, colorNext }) => {
+  let previousStatus: StepProps['status'];
+  return (
+    <StyledUl>
+      {steps.reverse().map((step, index) => {
+        const { date, text, status, button } = step;
+        const statusIcon = getStatusIcon(status);
+        if (index > 0 && steps.length > 1) {
```
edenvin, line 98: Will `steps.length` ever not be > 1 if index > 0?

src/Molecules/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,134 @@
...
+          previousStatus = steps[index - 1].status;
+        }
+        return (
+          <StyledListItem
+            /* eslint-disable-next-line react/no-array-index-key */
+            key={`${date}_${index}`}
+            previousStatus={previousStatus}
```
edenvin, line 105: From what I can see, this is the only place where `previousStatus` is used, so I think you should be able to replace it with this
```
steps[index - 1]?.status
```
If we are on index = 0 the expression will return undefined, else it will return `status`.

src/Molecules/Timeline/Timeline.tsx
```diff
@@ -0,0 +1,134 @@
...
+    width: 2px;
+    z-index: 1;
+    background: ${(() => {
+      switch (status) {
+        case 'ACTIVE':
+        case 'FAILURE':
+          return colorNext ? colorNext(theme) : theme.color.timelineNext;
+        case 'SUCCESS':
+        default:
+          return colorSuccess ? colorSuccess(theme) : theme.color.timelineSuccess;
```
KitsuneKelso, line 53: There seems to be something wrong with the colors of the lines, shouldn't the completed lines be blue and the failed lines be red?
Screenshot 2020-06-30 at 08 46 40

edenvin: Jesper and I talked about that. In the designs it seems to be green lines between green circles, and gray otherwise. It's a bit unclear, though, but I thought this looked better also.

KitsuneKelso: Alright, then I'll change to an approved. 🙂

src/Molecules/Timeline/Timeline.tsx
anneback: prettier didn't like this :(.

src/Molecules/Timeline/Timeline.tsx
edenvin: Did you forget this? :)

PR #607, opened 2020-06-30
Navtabs no router

JohnPhoto:
NavTabs and it's children no longer assumes that react-router is
available but instead accepts active prop and lets consumer decide if
active or not.

BREAKING CHANGE: 🧨 Dropping support for automatic active with react-router


JohnPhoto:
Will change this PR to beta when beta is up to date. See #611

PR #610, opened 2020-07-02
Feat/round slider 2

wiktoriatomzik:


src/Molecules/Slider/Slider.stories.tsx
```diff
@@ -26,6 +26,7 @@ export const Default = () => {
           onChange={handleChange}
           value={value}
           {...getSliderProps({ min: 0, max: 100, step: 1 })}
+          sliderColor={t => t.color.pill1}
```
mjmonline, line 29: There is no default color?

src/Molecules/Slider/Slider.tsx
```diff
@@ -1,82 +1,121 @@
...
   );
-  height: 15px;
+  background: ${p => p.$disabled && p.theme.color.sliderDisabled};
+  ${height};
   max-width: 100%;
   width: 100%;
 `;
 
-const StyledSlider = styled.div`
+const StyledSlider = styled('div')<InternalSliderTypes>`
```
mjmonline, line 64: why not `styled.div`?

src/Molecules/Slider/Slider.tsx
```diff
@@ -1,82 +1,121 @@
...
   transform: translateY(-50%);
-  background: ${p => p.theme.color.sliderThumbColor};
-  border: 1px solid ${p => p.theme.color.sliderThumbBorder};
+  background: ${p => p.theme.color.sliderThumbBackground};
+  border: ${p =>
+    `${p.theme.spacing.unit(1)}px solid ${
+      p.$sliderColor && !p.$disabled ? p.$sliderColor(p.theme) : p.theme.color.sliderColor
+    }`};
+  border: ${p =>
+    p.$disabled ? `${p.theme.spacing.unit(1)}px solid ${p.theme.color.sliderDisabled}` : ''};
```
mjmonline, line 97: if it's disabled you will set border to empty here, also it comes after the rule on line 92 so it's more important, bug potential

src/Molecules/Slider/Slider.tsx
```diff
@@ -159,53 +200,65 @@ const Slider: FC<Props> = ({ onChange, value, max, min, step, leftColor, rightCo
...
+        if (typeof onChange === 'function') {
+          onChange(newX);
+        }
       }
     }
   };
 
-  const gradient = {
+  const sliderBackground = {
     background: `linear-gradient(to right, ${
```
mjmonline, line 227: you have the same styles on `StyledSliderWrapper` any reason?

src/Molecules/Slider/Slider.stories.tsx
wiktoriatomzik: there is, check line 167

src/Molecules/Slider/Slider.tsx
wiktoriatomzik: if it is not disabled* and then the previous rule will apply.

src/Molecules/Slider/Slider.tsx
wiktoriatomzik: I forgot ;_;

src/Molecules/Slider/Slider.tsx
wiktoriatomzik: the procent of linearGradient is recalculated on every thumb change (with the getPercentage function). if you disable any of them, the styles don't change dynamically (I've tried it)

wiktoriatomzik:
> Your components height is equal to the slider bar height. This makes it tricky to put in a layout because designers adds spacing to surrounding components based on the whole height of the slider (including thumb height). So for example when a developer sees 10px padding in design and tries to apply that he/she will actually just have 2px spacing

@mjmonline, any suggestions on how to solve that?

src/Molecules/Slider/Slider.stories.tsx
mjmonline: Then in the default story we try to use all default values, so you can remove `sliderColor` from this story

mjmonline:
> > Your components height is equal to the slider bar height. This makes it tricky to put in a layout because designers adds spacing to surrounding components based on the whole height of the slider (including thumb height). So for example when a developer sees 10px padding in design and tries to apply that he/she will actually just have 2px spacing
> 
> @mjmonline, any suggestions on how to solve that?

@wiktoriatomzik you can have margins on `StyledSlider` to make it equal the height of the thumb

wiktoriatomzik:
> > > Your components height is equal to the slider bar height. This makes it tricky to put in a layout because designers adds spacing to surrounding components based on the whole height of the slider (including thumb height). So for example when a developer sees 10px padding in design and tries to apply that he/she will actually just have 2px spacing
> > 
> > 
> > @mjmonline, any suggestions on how to solve that?
> 
> @wiktoriatomzik you can have margins on `StyledSlider` to make it equal the height of the thumb

@mjmonline would love to do it like that. this is the outcome:
image

src/Molecules/Slider/Slider.stories.tsx
wiktoriatomzik: done 👌

PR #649, opened 2020-08-07
Mobile multistep progress

kmelkon:


src/Molecules/MultiStepProgress/TopLevel/TopLevel.tsx
```diff
@@ -0,0 +1,233 @@
...
+        titleDone={titleDone}
+        titleNotDone={titleNotDone}
+        isInDrawer
+      />
+    );
+  }
+
+  return (
+    <>
+      <Media query={t => t.media.greaterThan(t.breakpoints.md.size)}>
```
nn-davsko, line 172: Just out of curiosity, how come you use ...md.**size** here and not at line 182?

mjmonline: it's not my code, I think @kmelkon was just testing it out :)

PR #651, opened 2020-08-13
Fix/table key prop

fredriklu:
The new FlexTable missed som key-props for some maps, which this fixes

### DIFF
--- a/src/Molecules/FlexTable/shared/RenderForSizes/RenderForSizes.tsx
+++ b/src/Molecules/FlexTable/shared/RenderForSizes/RenderForSizes.tsx
@@ -62,7 +62,7 @@ export const RenderForSizes: RenderForSizesComponent = ({
         const container = Container({ children: component, ...props });
 
         if (size === 'xs' && !nextSize) {
-          return container;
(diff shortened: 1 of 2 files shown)


src/Molecules/FlexTable/shared/RenderForSizes/RenderForSizes.tsx
```diff
@@ -62,12 +62,18 @@ export const RenderForSizes: RenderForSizesComponent = ({
...
-          return container;
+          return <React.Fragment key={size}>{container}</React.Fragment>;
         }
 
         const as = (() => container) as React.FC;
 
-        return <Media key={size} query={t => getMediaQuery(t, size, nextSize)} as={as} />;
+        return (
+          <Media
+            key={`${size}${index + 1}`}
```
toblu, line 72: Do we want to keep `index` as part of the key here, or should we align it with the key on line 65 using just `size`? Using only size should be enough since it will be unique among its direct siblings right?

nn-davsko: We should probably get rid of it like you say

PR #657, opened 2020-08-19
feat: 🎸 add csv svg to icon component

robinandersson:


gustav-pettersson-nordnet (approved):

oscarsandberg (approved):

src/Atoms/Icon/components/Csv.tsx
```diff
@@ -0,0 +1,30 @@
...
+const StyledPath = styled.path<StyledChildProps>`
+  ${p => {
+    const strokeColor = getColor(p.theme, p.theme.color.svgStokeLight, p.strokeColorFn);
+    return `fill: ${strokeColor};`;
+  }}
+`;
+
+export const Csv = ({ stroke, ...props }: ChildProps) => {
+  return (
+    <IconBase {...props} viewBox="0 0 16 16">
```
nn-davsko, line 16: did you pass this through any compression tool like svgomg? I can't tell if it's right or wrong but by the looks of it you have a lot of information here, but perhaps the ending result looks like this for this icon?

nn-davsko (commented):
Looks good, please have a look at my question in Csv.tsx

src/Atoms/Icon/components/Csv.tsx
nn-davsko: note: comment relates more to whats inside <IconBase> than this actual line.

gustav-pettersson-nordnet: We did run it through svgomg with the standard setting and use of view port instead of pixels. Might be some setting to reduce it even further.

PR #661, opened 2020-08-24
fix: 🐛 prevent unnecessary remounts on initial flextable render

toblu:
The change in IsomorphicMedia fixes a bug that caused FlexTable cells to be remounted multiple times on initial render.

The render condition for the CSS fallback in SSR was too broad and also included the first render cycle in the client. Added an extra check to only render the CSS fallback on the server.

Using IsomorphicMedia in RenderForSizes also solves the flickering issue when expanding collapsed row content. See #650


src/Atoms/IsomorphicMedia/IsomorphicMedia.tsx
```diff
@@ -17,8 +17,9 @@ const StyledDiv = styled.div<{ query: Props['query'] }>`
   }
 `;
 
-const useIsomorphicLayoutEffect =
-  typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;
+const isServerSide = typeof window === 'undefined';
```
smithsamyan, line 20: Tip: instead of defining this yourself, you can rely on the useSSR hook

smithsamyan (approved):

gustav-pettersson-nordnet (approved):

src/Atoms/IsomorphicMedia/IsomorphicMedia.tsx
toblu: Good tip, thanks! I will update

nn-davsko (approved):

PR #678, opened 2020-09-07
fix: add possibility to override with zero gutter

oscarsandberg:
Now possible to override with zero gutter for different screen sizes. Also add story to show the functionality.


almightybruce (approved):
sublime

toblu (approved):

robinandersson (approved):

nn-davsko (approved):
Nice, also looks like we need to have a look at what formatting rules we are using.

gustav-pettersson-nordnet (approved):

gustav-pettersson-nordnet:
> Nice, also looks like we need to have a look at what formatting rules we are using.

It's prettier that has been updated. Should match other repositories now.

PR #679, opened 2020-09-07
Feat/export expand row

robinandersson:
Export ExpandRow from FlexTable


almightybruce (approved):

oscarsandberg (approved):

gustav-pettersson-nordnet (approved):

smithsamyan (commented):
Question: should we update the virtualisation example with the new implementation? Isnt that the reason for this export?

nn-davsko (approved):

robinandersson:
> Question: should we update the virtualisation example with the new implementation? Isnt that the reason for this export?

Good point,

the main reason for the export is to enable us to implement virtualisation for tables with expanded row, yes. We need a new version (with this export) to continue other work. Either we merge this now and update the examples later, or we update the example straight away.

PR #681, opened 2020-09-08
Feat/checkbox sizes

nn-davsko:
There's designs for 16x16 sized checkboxes (in this PR size 's'), also MultiSelectList in Input.Select contains a snowflake checkbox. We just thought it would be better to have medium and small sized checkboxes.


jesperwiner (approved):

KitsuneKelso (approved):

almightybruce (approved):
we should really run prettier on all of ui to avoid this many unrelated changes

PR #683, opened 2020-09-08
Fix/multistep border color and new thin icons

LifeofMichal:
Mutlistep progress bar margins between steps, and smaller margins on mobile.
Plus some new icon variants for page-mortgage-application.

The spaces are transparent, but not visible on Norndet/UI because of the div.root, only present on UI.
Screenshot 2020-09-08 at 14 44 34


src/Atoms/Icon/components/TrashRemove.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { IconBase } from '../IconBase';
+import { BaseProps } from '../IconBase.types';
+
+export const TrashRemove = (props: BaseProps) => {
```
jesperwiner, line 5: replace existing Trash instead of adding a new

jesperwiner (changes requested):

src/Atoms/Icon/components/CrossCircleThin.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { ChildProps, StyledChildProps } from '../IconBase.types';
+
+const StyledPolygon = styled.polygon<StyledChildProps>`
+  ${(p) => {
+    const strokeColor = getColor(p.theme, p.theme.color.svgStokeLight, p.strokeColorFn);
+    return `fill: ${strokeColor};`;
+  }}
+`;
+
+export const CrossCircleThin = ({ stroke, ...props }: ChildProps) => {
```
jesperwiner, line 13: its just a black circle on the demo page

src/Atoms/Icon/components/TrashRemove.tsx
jesperwiner: also the designers said that they want the existing ones to be replaced with the new thin variants, no need to add more icons, just replace the svg path in the existing

victoriatomzik (approved):

jesperwiner (approved):

PR #685, opened 2020-09-08
fix: add contain prop to solve ie11 wrapping issue with flexbox

mjmonline:


src/Atoms/Flexbox/Flexbox.stories.tsx
```diff
@@ -209,3 +208,30 @@ export const withHeightModified = () => (
 withHeightModified.story = {
   name: 'With height modified',
 };
+
+export const withContain = () => (
```
smithsamyan, line 212: Tip: Worth showing one without contain so people can see the difference in the browser

smithsamyan (approved):

PR #692, opened 2020-09-15
Fix/update filter icon

gustav-pettersson-nordnet:


almightybruce:
looks like some of the caption-work went into this, is that intended?

gustav-pettersson-nordnet:
other commits sliped in.

PR #694, opened 2020-09-17
Feature/index funds colors

KitsuneKelso:
Not sure why the Flexbox storyshot got updated, maybe someone forgot to update it when they removed something. 🤷‍♂️

### DIFF
--- a/src/theme/Theme.stories.tsx
+++ b/src/theme/Theme.stories.tsx
@@ -2,25 +2,42 @@ import React from 'react';
 import styled from 'styled-components';
 import MD from 'react-markdown';
 import { propOr } from 'ramda';
-import { theme, createTheme, Table, Thead, Tbody, Th, Tr, Td } from '..';
+import { theme, createTheme, Table, Thead, Tbody, Th, Tr, Td, Flexbox } from '..';
 import { rawColor } from './theme';
 import colorDocs from './Colors.md';
 import { Display } from '../common/Display';
 
 const Color = styled.div`
-  width: ${p => p.theme.spacing.unit(14)}px;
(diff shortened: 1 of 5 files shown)


src/theme/theme.ts
```diff
@@ -241,6 +241,13 @@ const createColors = (type: ThemeColorsVersion): ThemeColors<typeof type> => {
     progressBarDone: rawColor.positive,
     progressBarActive: rawColor.cta,
     progressBarNext: rawColor.gray4,
+    indexFundsBackground: rawColor.gray6,
+    indexFundsNorwegianAccent: rawColor.brandBlue,
+    indexFundsFinnishAccent0: rawColor.brandPink,
```
jesperwiner, line 246: would you mind if I said I prefer having these in an array instead like the graph ones I just changed ?

KitsuneKelso: I could try that out, if that's the preferred way now. 🙂

src/theme/Theme.stories.tsx
```diff
@@ -2,25 +2,42 @@ import React from 'react';
...
-  </>
-);
+const ColorInArray = styled.div`
+  width: ${(p) => p.theme.spacing.unit(4)}px;
+  height: ${(p) => p.theme.spacing.unit(4)}px;
+  padding: 0;
+  background-color: ${(p) => p.color};
+  border: 1px solid #eee;
+  display: ;
+`;
```
KitsuneKelso, line 25: Here's how this will look now in the story:
Screenshot 2020-09-17 at 11 21 08

jesperwiner (approved):

anneback (approved):

PR #695, opened 2020-09-17
Feat/tax icon

fredriklu:


nn-davsko (approved):

almightybruce:
just a thought: is it possible to convert the text to a path so we don't need to load our font first?

almightybruce (approved):

PR #700, opened 2020-09-21
Feat: Datepicker

frozzare:


src/Molecules/Datepicker/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+const StyledCalendarDay = styled(Box)`
+  background: ${({ theme }) => theme.color.backgroundInput};
+  min-width: 38px;
+  min-height: 38px;
+  border: 1px solid transparent;
+  justify-content: center;
+  align-items: center;
+  display: flex;
+  color: gray;
```
robineng, line 27: Maybe use a theme color

src/Molecules/Datepicker/SelectYear/SelectYear.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { useMemo, useState } from 'react';
+import styled from 'styled-components';
+import { Props } from './SelectYear.types';
+import { Box, Input, Flexbox, Icon, Typography } from '../../..';
+import { newDate } from '../shared/dateUtils';
+
+const StyledInputSelect = styled(Input.Select)`
+  > div > div > div {
```
robineng, line 8: ⏩

src/Molecules/Datepicker/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,160 @@
...
+      onClick(arrowDate);
+    }
+  }, [arrowDate, arrowLeft, arrowRight, arrowUp, arrowDown, onClick]);
+
+  const calendar = getCalendar(now, {
+    locale: getLocale(locale),
+  });
+
+  return (
+    <Flexbox container direction="column" data-testid="datepicker-calendar">
```
iamstarkov, line 131: so, what happens when there are two calendars on a page?

frozzare: The calendar component is used by the Datepicker component so you won't use the Calendar page when having multiple datepickers on a page.

iamstarkov (approved):

src/Molecules/Datepicker/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,160 @@
...
+    isToday(date) ? 'today' : '',
+  ];
+
+  const handleOnClick = useCallback(() => {
+    if (disabled) {
+      return;
+    }
+
+    if (onClick) {
+      onClick(date);
```
Dhq, line 83: Should the onClick be returned, so in the cases where it is a promise it is would be possible for consumers to await it?

src/Molecules/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,162 @@
...
+const StyledDropdownBubble = styled(DropdownBubble)`
+  max-width: 301px;
+  margin-top: -11px;
+  &:after,
+  &:before {
+    display: none;
+  }
+`;
+
+// different version from the one in the icon components.
```
Dhq, line 29: Feels like it should be placed with the other icons anyway https://github.com/nordnet/ui/tree/master/src/Atoms/Icon/components

src/Molecules/Datepicker/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,160 @@
...
+const StyledCalendarDay = styled(Box)`
+  background: ${({ theme }) => theme.color.backgroundInput};
+  min-width: 38px;
+  min-height: 38px;
+  border: 1px solid transparent;
+  justify-content: center;
+  align-items: center;
+  display: flex;
+  color: ${({ theme }) => theme.color.label};
+  margin-bottom: 2px;
```
Dhq, line 28: I would have a `cursor: pointer` on this one, it feels weird selecting stuff with a click with the arrow- or text cursor. What do you think?

src/Molecules/Datepicker/Calendar/Calendar.tsx
Dhq: Ah, this onClick comes from within the Datepicker and not from the consumer. Then never mind

src/Molecules/Datepicker/Datepicker.tsx
frozzare: Yeah, i have tough about this. I'll move it and just calling `CalendarTwoDots` or something 👍

src/Molecules/Datepicker/Calendar/Calendar.tsx
frozzare: Ah, yes that's a good catch!

jesperwiner:
Do we support different locales ? When changing the locale in storybook nothing happens to date format.

nn-davsko:
Does anyone feel qualified to verify the a11y requirements on this component? I've navigated through most of the parts using voice over but I am not sure what to expect.

frozzare:
> Do we support different locales ? When changing the locale in storybook nothing happens to date format.

We do, but the format isn't update right now. If you open the calendar you can see the locale working. Will look into the format 👍

src/Molecules/Datepicker/Datepicker.tsx
```diff
@@ -0,0 +1,146 @@
...
+        setSelectedDate(date);
+      }
+    },
+    [setSelectedDate],
+  );
+
+  const handleInputOnFocus = useCallback(() => setOpen(true), [setOpen]);
+
+  const datepicker = (
+    <Box my={3} mx={2.8}>
```
mjmonline, line 101: full units

src/Molecules/Datepicker/Header/Header.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Props } from './Header.types';
+import { Box, Flexbox, Link, Icon } from '../../..';
+import SelectMonth from '../SelectMonth';
+import SelectYear from '../SelectYear';
+
+const Header: React.FC<Props> = ({ id, locale, now, onMonthChange, onYearChange }) => {
+  return (
+    <Flexbox container justifyContent="space-between">
+      <Flexbox item>
+        <Box mt={0.8} ml={-0.1}>
```
mjmonline, line 11: full units

mjmonline:
Some a11y comments:
1. First thing I reacted to was that the color used for marking todays date is really hard to see even for me with perfect vision.
2. I tested the gray colors against each other when having disabled dates and they fail AA requirement, maybe design is aware of this and maybe not.
3. Would be nice to enter date manually, users can see the pattern
4. When opening the datepicker I would like a day to be focused like todays date or the previously chosen date
4. Forward and backwards arrows just reads "button" which is of no help for a blind user, would be nice if they could read "previous month september 2020, button" and "next month december 2020, button". Something like that
5. Wrap the weekdays row in `aria-hidden` because stepping through them and getting "mon, tue, wed, thu, fri, sat, sun" and not being able to do anything with them gives nothing to the user.
6. When stepping through the dates I hear 5, 6, 7, 8 would be helpful if I instead got a readable date like "thursday 5th of october, 2020"

src/Molecules/Datepicker/SelectMonth/SelectMonth.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const months = [...Array(12).keys()];
+
+const StyledInputSelect = styled(Input.Select)`
+  margin-right: 12px;
+
+  > div > div > div {
+    margin-right: -22px;
+    margin-top: -10px;
+  }
```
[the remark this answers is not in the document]
frozzare: Yeah, this is a bit weird. The arrow icon of the input select and the arrow icons of select month/year should be aligned and it's not possible to get it align using theme units unfortunately.

frozzare:
So a update what's fixed from the a11y feedback: 

> First thing I reacted to was that the color used for marking todays date is really hard to see even for me with perfect vision.

This is now fixed, it was the wrong colors.

> I tested the gray colors against each other when having disabled dates and they fail AA requirement, maybe design is aware of this and maybe not.

This is now fixed, it was the wrong colors.

> Would be nice to enter date manually, users can see the pattern

Yes, I agree, this is now fixed and if you write the date in the right format order it will be parsed and the ui will update to that date.

> Forward and backwards arrows just reads "button" which is of no help for a blind user, would be nice if they could read "previous month september 2020, button" and "next month december 2020, button". Something like that

Added aria label text and props to translate it.

> Wrap the weekdays row in aria-hidden because stepping through them and getting "mon, tue, wed, thu, fri, sat, sun" and not being able to do anything with them gives nothing to the user.

Wrapped this in aria hidden attribute.

> When stepping through the dates I hear 5, 6, 7, 8 would be helpful if I instead got a readable date like "thursday 5th of october, 2020"

Added aria label for this text.

> When opening the datepicker I would like a day to be focused like todays date or the previously chosen date

This isn't fixed since we would move the focus away from the input field and then you can't enter date manually, if it possible to solve so you can write in the input field and have focus on a date we could do something. But not sure how.

Dhq (approved):

PR #702, opened 2020-09-23
Feature/add pagination component

toblu:
Add Pagination component


nn-davsko (approved):
Looks great, well done!

src/Molecules/Pagination/DesktopPagination.tsx
```diff
@@ -0,0 +1,73 @@
...
+const CHEVRON_PADDING = 2 * 8;
+const BETWEEN_8_AND_13_PAGES_WIDTH = 150;
+const BETWEEN_14_AND_101_PAGES_WIDTH = 168;
+const MORE_THAN_101_PAGES_WIDTH = 200;
+
+const StyledFlexbox = styled(Flexbox)<{ numberOfPages: number }>`
+  width: ${(p) => {
+    if (p.numberOfPages < 8) return p.numberOfPages * NUMBER_WIDTH + CHEVRON_PADDING;
+
+    if (p.numberOfPages >= 8 && p.numberOfPages >= 13) return BETWEEN_8_AND_13_PAGES_WIDTH;
```
fredriklu, line 17: It seems like the true width of this example in the story is actually 200px wide? How does it come that this default width is larger than the 999 example on start? (If you refer to the distance between the chevrons). 

Also, the 999 example increases in size as the width of the numbers increase (e.g. if you navigate to page 996), but it doesn't for the other examples. Shouldn't all examples be adaptive if possible?

src/Molecules/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const paginationWithManyPages = () => (
+  <PaginationWrapper totalItems={999} itemsPerPage={1} />
+);
+
+paginationWithManyPages.story = {
+  name: '999 pages',
+};
+
+export const paginationCompactMobile = () => (
+  <PaginationWrapper totalItems={10} itemsPerPage={1} compact />
```
fredriklu, line 70: This story redirects weirdly in the vercel-preview, just wanted to let you know!

fredriklu (approved):

fredriklu:
Great that you've added documentation right off the bat! 🌟

src/Molecules/Pagination/DesktopPagination.tsx
toblu: Really good catch! That is of course not the way it should be, something must have happened here... Looking into it now

toblu: Ah, it's the `p.numberOfPages >= 13` that is the culprit here. Since we want to check if it's between 8 and 13 pages it should ofc be `p.numberOfPages <= 13` 🙈 . I will fix that!

mhogler (approved):

src/Molecules/Pagination/Pagination.stories.tsx
toblu: How do you mean? It looks fine from what I can tell

src/Molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,64 @@
...
+  itemsPerPage,
+  totalItems,
+  compact,
+  onPageChange,
+}) => {
+  const numberOfPages = Math.ceil(totalItems / itemsPerPage) || 1;
+
+  const onClickPageItem = useCallback(
+    (e) => {
+      onPageChange(Number(e.target.textContent));
```
smithsamyan, line 19: Tip: It would be better if you send the actual page number rather than using e.target.textContent. TextContent may become inaccurate if we have any formatting of the text or add visually hidden data for a11y. This wouldn't for example work if we had thousand separators

toblu: Yes, that is a good point 👍  Will update

src/Molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,64 @@
+import React, { useCallback } from 'react';
+import { PaginationProps } from './Pagination.types';
+import DesktopPagination from './DesktopPagination';
+import Media from '../../Atoms/Media';
+import MobilePaginationCompact from './MobilePaginationCompact';
+import MobilePaginationDefault from './MobilePaginationDefault';
+
+const Pagination: React.FC<PaginationProps> = ({
+  currentPage,
```
smithsamyan, line 9: tip: In the ui library, we try  to implement both controlled and uncontrolled behaviours. You don't have to do it now but we should implement pagination with uncontrolled behaviour :)

src/Molecules/Pagination/Pagination.tsx
toblu: Fixed in d5bc24e

src/Molecules/Pagination/MobilePaginationDefault.tsx
```diff
@@ -0,0 +1,112 @@
...
+        onClick();
+      }
+    }}
+  >
+    <StyledBox variant={variant}>{children}</StyledBox>
+  </StyledLink>
+);
+
+const StyledFlexbox = styled(Flexbox)<{ numberOfPages: number }>`
+  width: ${(p) => Math.min(p.numberOfPages, MAX_NUMBER_ITEMS) * ITEMS_TOTAL_WIDTH}px;
```
mjmonline, line 57: kinda bug prone, is there a reason to set the width here? Can it be solved with just paddings?

src/Molecules/Pagination/MobilePaginationDefault.tsx
```diff
@@ -0,0 +1,112 @@
...
+const StyledLink = styled.a`
+  justify-content: inherit;
+  text-decoration: none;
+  color: inherit;
+`;
+
+const StyledBox = styled(Box)<{ variant?: string }>`
+  display: flex;
+  height: 40px;
+  width: 40px;
```
mjmonline, line 19: units

src/Molecules/Pagination/MobilePaginationCompact.tsx
```diff
@@ -0,0 +1,59 @@
...
+  <StyledBox
+    href="#"
+    role="button"
+    onClick={onClick}
+    onKeyDown={(e) => {
+      // Link should trigger on spacebar clicked like actual button.
+      if (e.keyCode === 32) {
+        onClick();
+      }
+    }}
```
mjmonline, line 29: same here, why not a `<button>`?

src/Molecules/Pagination/MobilePaginationCompact.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Flexbox, Icon } from '../..';
+import { PaginationCompactProps, BrowseButtonProps } from './Pagination.types';
+
+const StyledBox = styled.a`
+  display: flex;
+  height: 40px;
+  width: 40px;
```
mjmonline, line 9: units

src/Molecules/Pagination/DesktopPagination.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+const StyledFlexbox = styled(Flexbox)<{ numberOfPages: number }>`
+  width: ${(p) => {
+    if (p.numberOfPages < 8) return p.numberOfPages * NUMBER_WIDTH + CHEVRON_PADDING;
+
+    if (p.numberOfPages >= 8 && p.numberOfPages <= 13) return BETWEEN_8_AND_13_PAGES_WIDTH;
+
+    if (p.numberOfPages >= 14 && p.numberOfPages <= 101) return BETWEEN_14_AND_101_PAGES_WIDTH;
+
+    return MORE_THAN_101_PAGES_WIDTH;
```
mjmonline, line 21: see my previous comment on width

src/Molecules/Pagination/DesktopPagination.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }}px;
+`;
+
+const PageNumberItem: React.FC<PageItemProps> = ({ active = false, children, onClick }) => (
+  <Flexbox item>
+    <Button id="currentPage" variant="neutral" disabled={active} onClick={onClick}>
```
mjmonline, line 27: you're not using the id but you have several buttons now with same id

mjmonline: Also in mobile you use links, whats the difference?

mjmonline (changes requested):
So in this repo we try to keep high standard and we try to follow the "DND Definition of done checklist" in confluence, message me or any of the #ui-task-force members for the link.

So about buttons and links, if we are redirecting the user to let's say "page2" then the logical solution would be to use a link with the href="?page='2'" or something. If the pagination on the other hand is used for a let's say sub section of a page and it's not really considered as a "page2" then you're not redirecting so a button would be fine.

Also I'm not sure this PR has been tested with a screen reader because the result was not so user friendly, will post some videos explaining

mjmonline:
This is recording with VoiceOver on mac https://streamable.com/49ae9s

Firstly you should wrap the pagination in a `<nav role="navigation">` and label that element with for example `aria-label` so the user knows what they will be navigating.

Secondly your back and forward button just reads "button", it could be impossible for a visually impaired person to know what's going to happen if the button is clicked. But for a normal user he would just see the arrows and figure it out. Here I would suggest you use `<VisuallyHidden>` atom and write something like "go to previous page".

Thirdly the numbered buttons also just say "button 2", "button 3" and so on. Rather you would like it to read "go to page 2" or "go to page 3"

mjmonline:
This is recording of a old Pagination component in our old ui-kit: https://streamable.com/ykwijm

https://github.com/nordnet/nordnet-ui-kit/tree/master/src/components/pagination

src/Molecules/Pagination/Pagination.tsx
toblu: Fixed in cd1ba98

src/Molecules/Pagination/__snapshots__/Pagination.stories.storyshot
```diff
@@ -0,0 +1,3712 @@
...
+    display: none;
+  }
+}
+
+<div
+    className="c0"
+  >
+    <div
+      className="c1 c2"
+      numberOfPages={1}
```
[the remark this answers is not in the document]
toblu: Thanks, good catch. Fixed in d7dbfae

src/Molecules/Pagination/MobilePaginationDefault.tsx
mhogler: The reason why we must calculate the width is so that we can correctly position the chevron buttons from the beginning, otherwise they move as the user browses through the pages.

Imagine you're on page page 1 out of 12, and then you're stepping to page 4. The chevron buttons should stay in the same position even though the total width of the page items increases on page 4. Hopfully the screenshots makes this more clear.

We'd also like to make this less hard coded, but could not find any other way of solving it with paddings etc, do you have another suggestion of how it could be solved?
Screenshot 2020-09-24 at 14 43 41
Screenshot 2020-09-24 at 14 44 19

src/Molecules/Pagination/MobilePaginationDefault.tsx
mhogler: fixed in 93b6de8

src/Molecules/Pagination/MobilePaginationCompact.tsx
mhogler: fixed in 93b6de8

src/Molecules/Pagination/MobilePaginationDefault.tsx
```diff
@@ -0,0 +1,112 @@
...
+      if (variant !== 'selected' && onClick) {
+        onClick();
+      }
+    }}
+    onKeyDown={(e) => {
+      // Link should trigger on spacebar clicked like actual button.
+      if (e.keyCode === 32 && variant !== 'selected' && onClick) {
+        onClick();
+      }
+    }}
```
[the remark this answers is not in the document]
toblu: Good point, it should be a `<button>`. Fixed in 3024c70

src/Molecules/Pagination/MobilePaginationCompact.tsx
toblu: Fixed in 3024c70

src/Molecules/Pagination/DesktopPagination.tsx
toblu: Fixed in 3024c70

src/Molecules/Pagination/Pagination.md
```diff
@@ -0,0 +1,38 @@
+# Pagination
+
+This component enables the user to select a specific page from a range of pages. The number of pages is calculated based on the `totalItems` and `itemsPerPage` props.
+
+## Props
+
+| name         | type                   | default | description                      |
+| ------------ | ---------------------- | ------- | -------------------------------- |
+| currentPage  | number                 | N/A     | The current page number          |
```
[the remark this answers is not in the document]
mhogler: Fixed in a9617b6

toblu:
@mjmonline Thank you for your thorough review and great inputs! We have addressed all your comments, please take a look again :)

src/Molecules/Pagination/DesktopPagination.tsx
```diff
@@ -0,0 +1,90 @@
...
+    if (p.numberOfPages >= 14 && p.numberOfPages <= 101) return BETWEEN_14_AND_101_PAGES_WIDTH;
+
+    return MORE_THAN_101_PAGES_WIDTH;
+  }}px;
+`;
+
+const StyledList = styled.ul`
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
```
mjmonline, line 32: This is `Atom/List`

src/Molecules/Pagination/DesktopPagination.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { Props as FlexBoxProps } from '../../Atoms/Flexbox/Flexbox.types';
+import PageItems from './PageItems';
+import { PaginationDefaultProps, PageItemProps, BrowseButtonProps } from './Pagination.types';
+
+const StyledFlexbox = styled(
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  ({ numberOfPages, ...rest }: { numberOfPages: number } & FlexBoxProps) => <Flexbox {...rest} />,
+)`
+  width: ${(p) => {
+    const NUMBER_WIDTH = p.theme.spacing.unit(4.5);
```
mjmonline, line 13: We don't really have any fractions of a unit in the UI, somewhere there has been a mistake because designers is also aware of the 4px unit. It's not a blocker, I'm just informing

src/Molecules/Pagination/MobilePaginationDefault.tsx
```diff
@@ -0,0 +1,131 @@
...
+  border: none;
+  padding: 0;
+  justify-content: center;
+  align-items: center;
+`;
+
+const StyledList = styled.ul`
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
```
mjmonline, line 43: same, can be reused

src/Molecules/Pagination/DesktopPagination.tsx
```diff
@@ -0,0 +1,90 @@
...
+  padding: 0;
+  list-style-type: none;
+`;
+
+const PageNumberItem: React.FC<PageItemProps> = ({ isCurrentPage = false, children, onClick }) => (
+  <Flexbox item as="li">
+    <Button
+      variant="neutral"
+      onClick={!isCurrentPage ? onClick : undefined}
+      aria-label={`${isCurrentPage ? 'Current Page:' : 'Go to'} page ${children}`}
```
mjmonline, line 40: Would be nice if devs can override the text with a prop and same for all other a11y texts

mjmonline (approved):

src/Molecules/Pagination/DesktopPagination.tsx
mhogler: Indeed, we thought about this as well, unfortunately with the current design, this was the only way we could make it work. Using either 4 or 5 units doesn't look good

PR #703, opened 2020-09-24
fix: enable keyboard keys when uncontrolled and adjust focus

mjmonline:


src/Molecules/Slider/Slider.tsx
```diff
@@ -72,6 +72,10 @@ const Container = styled.div<InternalProps>`
   align-items: center;
   height: ${(p) =>
     p.$variant === VARIANT_TYPES.SMALL ? `${THUMB_SMALL_PX}px` : `${THUMB_BIG_PX}px`};
+
+  &:focus {
+    outline: none;
```
victoriatomzik, line 77: is it accessible without focus?

mjmonline: yes, we have focus on the handle

victoriatomzik (approved):

PR #705, opened 2020-09-25
Avatar background color prop

johwer:


src/Molecules/Avatar/Avatar.tsx
```diff
@@ -14,17 +14,28 @@ const getSize = (p: StyledProps<Props>) => {
...
-export const Avatar: React.FunctionComponent<Props> = ({ children, size = 'm' }) => (
-  <StyledDiv container alignItems="center" justifyContent="center" size={size}>
-    <Typography type={size === 'm' ? 'tertiary' : 'caption'} color={t => t.color.textLight}>
+export const Avatar: React.FunctionComponent<Props> = ({
+  children,
+  size = 'm',
+  backgroundColors,
+}) => (
+  <StyledDiv
+    backgroundColors={backgroundColors}
```
edenvin, line 32: It is only one color, not multiple, so I think `backgroundColor` is a better name. Or, since it's a function that returns the color, maybe `getBackgroundColor` is even clearer?

johwer: Will change

gustav-pettersson-nordnet (approved):

edenvin (approved):

PR #706, opened 2020-09-29
Feat/custom tabs content

robinandersson:
Tab Content sets height to 100%, which is not compatible with e.g. putting tabs in the Drawer component (scroll appears).

This PR: pass Content as optional prop

Suggestions for other possible solutions are appreciated.

### DIFF
--- a/src/Molecules/Tabs/Tabs.tsx
+++ b/src/Molecules/Tabs/Tabs.tsx
@@ -123,28 +123,19 @@ export const Tabs: ContainerComponent = ({
         </Flexbox>,
       );
 
-      const tempProps = {
-        id: `tabs-tabpanel-${i}`,
-        role: 'tabpanel',
-        'aria-labelledby': `tabs-tab-${i}`,
-        hidden: !isActive,
-      };
-
-      const Temp = TabContentWrapper ? (
-        <TabContentWrapper {...tempProps}>{c}</TabContentWrapper>
-      ) : (
-        <Content
-          id={`tabs-tabpanel-${i}`}
-          role="tabpanel"
-          aria-labelledby={`tabs-tab-${i}`}
-          hidden={!isActive}
-        >
-          {c}
-        </Content>
-      );
-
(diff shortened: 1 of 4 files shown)


src/Molecules/Tabs/Tabs.tsx
```diff
@@ -123,15 +124,17 @@ export const Tabs: ContainerComponent = ({
...
-            {c}
-          </Content>
+        const tabContentProps = {
+          id: `tabs-tabpanel-${i}`,
+          role: 'tabpanel',
+          'aria-labelledby': `tabs-tab-${i}`,
+          hidden: !isActive,
+        };
+
+        contents = TabContentWrapper ? (
```
almightybruce, line 134: Looks good if you send something that can be JSX:ed, but will crash otherwise.. Perhaps some isElement check?

almightybruce: Didn't see the React.ComponentType typing. You can probably ignore the previous comment

src/Molecules/Tabs/Tabs.types.ts
```diff
@@ -23,6 +23,7 @@ export type ContainerProps = {
   children: React.ReactNode;
   className?: string;
   height?: number;
+  TabContentWrapper?: React.ComponentType;
```
toblu, line 26: If you want to also support intrinsic JSX strings like (`"div"`, `"span"` etc), this should be `React.ElementType` https://flow.org/en/docs/react/types/#toc-react-elementtype

robinandersson: Do we want to allow such elements though? I'm not sure either or.

toblu (approved):

nn-davsko (approved):
Where did we land on this one? Do it or scrap it?

robinandersson:
This PR is no longer needed or wanted.

To style the wrapper, use the approach introduced in #710

PR #710, opened 2020-09-30
Feat/export tabs content

gustav-pettersson-nordnet:


src/Molecules/Tabs/Tabs.stories.tsx
```diff
@@ -50,6 +50,30 @@ defaultStory.story = {
...
+    height: inherit;
+    background: ${({ theme }) => theme.color.sliderBackgroundColor};
+  }
+`;
+
+export const withStyledTabContent = () => (
+  <Typography type="secondary">
+    <StyledTabsContent>
+      <Tabs.Tab title="One" onTitleClick={action('Clicked title1')}>
+        Tab 1. Content have gray background
```
nn-davsko, line 64: Content has 🖌️

src/Molecules/Tabs/Tabs.tsx
```diff
@@ -73,13 +73,23 @@ const isItemOrUndefined = (x: any): x is { type: typeof Item; props: ItemProps }
...
+  TabContent,
+};
+
+export const Tabs: ContainerComponent & {
+  /**
+   * This will allow you to customize
+   * inner parts with styled-components
+   * @example
+   * const StyledContentTabs = styled(Tabs)`
+   *  & ~ ${Tabs.components.TabContent} {
```
nn-davsko, line 86: perhaps mention the selector for people who are not fluent in these types of things. like the "& ~" is siblings below, if it where a child it would just be ${Tabs.components.TabContent}.

I am not entirely sure about this, you could argue that the selector is something you should know or look up yourself depending on the scenario

nn-davsko: It's just because we got stuck here like 4 hours ...

src/Molecules/Tabs/Tabs.tsx
```diff
@@ -155,5 +165,7 @@ export const Tabs: ContainerComponent = ({
     </>
   );
 };
+
 Tabs.displayName = 'Tabs';
 Tabs.Tab = Item;
+Tabs.components = components;
```
nn-davsko, line 171: It is a little bit strange having Tab on Tabs while tab content is on Tabs.components.TabContent. I know the .components is a pattern here. Just food for thought and I dont think we want to break anything but
``` Tabs.Tab -> Tabs.Tab.TabContent or Tabs.components.Tab and Tabs.components.TabContent``` might make more sense.

nn-davsko (approved):
A few comments, nothing you need to adjust do with it as you like

nn-davsko (approved):

src/Molecules/Tabs/Tabs.tsx
robinandersson: Hehe yeah, that's true. Just a small FYI sentence about the reasoning wouldn't hurt

src/Molecules/Tabs/Tabs.tsx
robinandersson: We could probably either, although `Tabs.components.Tab` would require a breaking change (right?).

Also, I believe the reasoning for putting `TabContent` in a separate `components` is that they are not used directly/explicitly (as with the exported `Tab` component), i.e. they're only used to style them.

Simplest way to avoid confusion and avoid breaking change would probably be to reference `TabContent` as `Tabs.TabContent`.

fredriklu (approved):

PR #711, opened 2020-09-30
feat: 🎸 new style for mobile pagination compact

mhogler:
After discussion with designers we decided to make the compact pagination buttons for mobile slightly smaller


src/Molecules/Pagination/MobilePaginationCompact.tsx
```diff
@@ -3,26 +3,24 @@ import styled from 'styled-components';
...
   height: ${(p) => p.theme.spacing.unit(10)}px;
-  width: ${(p) => p.theme.spacing.unit(10)}px;
+  width: ${(p) => p.theme.spacing.unit(8)}px;
   border: none;
   padding: 0;
-  justify-content: center;
-  align-items: center;
   background-color: transparent;
   outline: none;
+  justify-content: ${(p) => (p.direction === 'left' ? 'flex-start' : 'flex-end')};
```
almightybruce, line 17: this is only valid if flex-direction is row, and not row-reverse.. right?

almightybruce: however maybe it's not possible to set flex-direction to anything other than row

mhogler: You're right, it's only valid for row. I added` flex-direction: row;`, just to be explicit. Thanks!

toblu (approved):

src/Molecules/Pagination/MobilePaginationCompact.tsx
gustav-pettersson-nordnet: make sure direction prop is not leaking out in the DOM.

gustav-pettersson-nordnet:
Will make a new one with the right title for new reviews.

PR #719, opened 2020-10-09
fix: 🐛 add quiet variant to input number

kmelkon:


src/Molecules/Input/Number/Number.tsx
```diff
@@ -171,11 +194,14 @@ const NumberInput: NumberComponent & {
     success,
     value: controlledValueRaw,
     visuallyEmphasiseRequired,
+    variant = 'normal',
   } = props;
   const [internalValue, setInternalValue] = useState(getNumberAsString(defaultValue));
   const intl = useIntl();
   const showSteppers = noSteppers !== true && isUndefined(leftAddon) && isUndefined(rightAddon);
```
jesperwiner, line 201: I would had added "&& variant !== 'quiet'" here instead and mot have the extra const and just pass variant below, can keep the comment to explain

src/Molecules/Input/Number/Number.types.ts
```diff
@@ -44,7 +45,7 @@ export type Props = {
   success?: boolean;
   value?: string | number;
   visuallyEmphasiseRequired?: boolean;
-  width?: string | number;
+  noSteppers?: boolean;
```
jesperwiner, line 48: just curious what prompted to move it here ?

src/Molecules/Input/Number/Number.stories.tsx
```diff
@@ -356,3 +356,17 @@ export const withPlaceholder = () => (
 withPlaceholder.story = {
   name: 'With placeholder',
 };
+
+export const withQuietVariant = () => (
```
jesperwiner, line 360: need exampes  to cover all the modes, check the other quiet variant

jesperwiner (changes requested):

src/Molecules/Input/Number/Number.types.ts
kmelkon: I was trying to make the types dependant on each other like if noSteppers is true then the variant can be quiet otherwise no but I failed. so this is just a remnant of that

src/Molecules/Input/Number/Number.tsx
kmelkon: That doesn't really work as you expect it to

victoriatomzik (approved):

jesperwiner (approved):

PR #733, opened 2020-10-20
FlexTable performance: prevent unnecessary mounting & unmounting of rows / cells

toblu:
This change fixes an issue that caused `RenderForSizes` to re-mount its children (`Container` & `Component` props) on every render. This causes a performance overhead in FlexTable since it means that all of the rows and cells are remounted multiple times.

We also changed the implementation of RenderForSizes to something that is easier to read. With the new implementation, RenderForSizes only renders children as a function with the correct props per screen size as argument.

Also renamed variables / functions / types in RenderForSizes to make it easier to understand


src/Molecules/FlexTable/BigFlexTable.stories.tsx
```diff
@@ -145,7 +145,7 @@ const FullHeightDiv = styled.div`
 export const VirtualizedTable = () => {
   const VirtualizedTableExample = () => {
     const ReactComponent = () => {
-      const rowsLength = number('Number of rows', 500);
+      const rowsLength = number('Number of rows', 100);
```
toblu, line 148: We reduced the default number of rows since 500 rows will make Storybook a sad panda 🐼 . However, if you eject the table component in an iframe, rendering 500 rows works fine

nn-davsko (approved):

nn-davsko (approved):

fredriklu (approved):

robinandersson (approved):

PR #736, opened 2020-10-22
feat: 🎸 implement status modal

crewneck:


src/index.ts
```diff
@@ -61,6 +61,7 @@ import SegmentedControl from './Molecules/SegmentedControl';
 import Select from './Molecules/Select';
 import SelectionCard from './Molecules/SelectionCard';
 import Slider from './Molecules/Slider';
+import StatusModal from './Molecules/StatusModal';
```
mjmonline, line 64: This component reuses other molecules, so the idea is that it should then be a organism. I'm not a fan of current structure but as long as we have it let's try to stick to it 🙂

src/Molecules/StatusModal/StatusModal.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+  .add('Loading', () => <StatusModal {...defaultProps} loading />)
+  .add('With all', () => <StatusModal {...defaultProps} status="CASE_1" options={options} />)
+  .add('Success', () => (
+    <StatusModal {...defaultProps} status="SUCCESS" options={options} />
+  ))
+  .add('Warning', () => (
+    <StatusModal {...defaultProps} status="WARNING" options={options} />
+  ))
+  .add('Error', () => (
+    <StatusModal {...defaultProps} status="ERROR" options={options} />
```
mjmonline, line 44: I think this API feels weird, so you pass in a object as `options` props and then with `status` prop you decide which of the `options` keys to use? In my opinion there is no need of letting the component handle this logic you could just pass the object you wanted in the first place like `options={options['SUCCESS']}` right?

Also this way of doing it devs have to always define the statuses for example that success needs to have the icon `<Icon.CheckMarkCircle color={(t) => t.color.positive} size={23} />` every time. And so on for the rest of the statuses

LifeofMichal: I found it weird too, but from what I've heard this is how Status Modal is implemented most of the time in UIM.

src/index.ts
LifeofMichal: That's a good point 👍🏻

src/Molecules/StatusModal/StatusModal.stories.tsx
crewneck: @mjmonline I think I see what you mean. I'll try to strip out more of what we send in to the component.

LifeofMichal (approved):

jesperwiner (approved):

PR #737, opened 2020-10-22
Fix/multistepprogress

victoriatomzik:


alvelid (approved):

kmelkon (approved):

src/Molecules/MultiStepProgress/TopLevel/TopLevel.tsx
```diff
@@ -51,31 +51,35 @@ const ListItem = styled.li<InternalProps>`
...
 
-  & + & {
-    ${({ theme }) => theme.media.lessThan(theme.breakpoints.md)} {
-      border-top: ${(p) => p.theme.spacing.unit(0.25)}px solid ${(p) => p.theme.color.divider};
-    }
-
-    ${({ theme }) => theme.media.greaterThan(theme.breakpoints.md)} {
-      margin-top: ${(p) => p.theme.spacing.unit(2)}px;
-    }
+const Wrapper = styled(Box)<InternalProps>`
```
Vikingjonsson, line 56: does this need InternalProps?

victoriatomzik: I just saw it myself and removed it ^_^

src/Molecules/MultiStepProgress/TopLevel/TopLevel.tsx
```diff
@@ -51,31 +51,35 @@ const ListItem = styled.li<InternalProps>`
     transition: opacity 0.16s ease-out;
     opacity: ${(p) => (p.$current ? 1 : 0)};
   }
+`;
 
-  & + & {
```
mjmonline, line 55: You probably don't want margin-top or border above the first item

src/Molecules/MultiStepProgress/TopLevel/TopLevel.tsx
```diff
@@ -51,31 +51,35 @@ const ListItem = styled.li<InternalProps>`
...
-      border-top: ${(p) => p.theme.spacing.unit(0.25)}px solid ${(p) => p.theme.color.divider};
-    }
-
-    ${({ theme }) => theme.media.greaterThan(theme.breakpoints.md)} {
-      margin-top: ${(p) => p.theme.spacing.unit(2)}px;
-    }
+const Wrapper = styled(Box)`
+  ${({ theme }) => theme.media.lessThan(theme.breakpoints.md)} {
+    margin: 0 ${({ theme }) => theme.spacing.unit(3)}px;
+    border-top: ${(p) => p.theme.spacing.unit(0.25)}px solid ${(p) => p.theme.color.divider};
```
mjmonline, line 59: I see that you just copy pasted but here use 1px instead of 0.25 units because this should not get bigger if our unit changes also we don't do fractions of a unit

src/Molecules/MultiStepProgress/TopLevel/TopLevel.tsx
```diff
@@ -51,31 +51,35 @@ const ListItem = styled.li<InternalProps>`
...
 
 const Content = styled(Box)`
   position: relative;
 `;
 
-const StyledButton = styled(Button)`
+const StyledButton = styled(Button)<InternalProps>`
   justify-content: flex-start;
+  ${({ theme }) => theme.media.lessThan(theme.breakpoints.md)} {
+    margin-left: ${(p) => (p.$isInDrawer ? -p.theme.spacing.unit(3) : 0)}px;
```
mjmonline, line 73: why is this?

mjmonline (changes requested):

src/Molecules/MultiStepProgress/TopLevel/TopLevel.tsx
mjmonline: There is no story with the multistep in a drawer, wanna create one?

Vikingjonsson (approved):

src/Molecules/MultiStepProgress/TopLevel/TopLevel.tsx
```diff
@@ -52,12 +47,24 @@ const ListItem = styled.li<InternalProps>`
...
+      left: ${({ theme }) => theme.spacing.unit(3)}px;
+      right: ${({ theme }) => theme.spacing.unit(3)}px;
     }
+  }
+`;
 
-    ${({ theme }) => theme.media.greaterThan(theme.breakpoints.md)} {
+const Wrapper = styled(Box)<InternalProps>`
+  ${({ theme }) => theme.media.greaterThan(theme.breakpoints.md)} {
+    li + li & {
```
mjmonline, line 67: This is a little dangerous because it says any `li`that has a direct sibling `li` that also has the `Wrapper` as a child. Which in this case works but this is a UI component and it could be used on different pages with different html. So keep it safe by doing `ListItem + ListItem` instead of `li + li`

mjmonline (changes requested):

mjmonline (approved):

PR #738, opened 2020-10-22
Fix/pagination maximum depth when controlled

fredriklu:


src/Molecules/Pagination/Pagination.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { useCallback, useEffect, useState } from 'react';
+import R from 'ramda';
```
robinandersson, line 2: With never versions of Ramda I believe it should be

`import * as R from 'rambda;`

robinandersson (approved):

smithsamyan (approved):

mhogler (approved):

PR #745, opened 2020-10-27
FlexTable: support flexprops only on header cells with SSR

toblu:
Bring back support for the old FlexTable API (setting flexprops only on the header cells) and make other cells and footers inherit the column settings, this time with SSR support! ⚡ 

The way this can be implemented is by using a JavaScript closure around the `getColumnFlexProps` function. Since the HeaderRow will be rendered first, it will call the `getColumnFlexProps` without any previously calculated flexprops, and then the result from calling `getColumnFlexProps` will be "stored in the closure" for that specific columnId. This means that when a cell with the same columnId calls the `getColumnFlexProps` function, the previously calculated flexprops from the Header can be returned directly. This removes the need for supplying flexprops for each cell, while maintaining SSR compatibility.


nn-davsko:
This is great but how are you sure the header row will be rendered before the other rows? Aren't they all siblings? I am sure you're right I'd just like to know how we can know this?

nn-davsko (commented):
Niiice

nn-davsko (approved):

almightybruce (approved):
nice

toblu:
> This is great but how are you sure the header row will be rendered before the other rows? Aren't they all siblings? I am sure you're right I'd just like to know how we can know this?

That is a very valid point. Intuitively it makes sense that components would be rendered in order from top to bottom given the normal execution order of JS, and when I have tested this it has always been the case that all the headers are rendered before the cells. 

However, I can't actually find anything in the docs that tells me that this will always be the case. I agree that it would feel better if we could find something that tells us that we can rely on this being the case. I will look around more to see if I can find some resources

fredriklu (approved):

mhogler (approved):

toblu:
> > This is great but how are you sure the header row will be rendered before the other rows? Aren't they all siblings? I am sure you're right I'd just like to know how we can know this?
> 
> That is a very valid point. Intuitively it makes sense that components would be rendered in order from top to bottom given the normal execution order of JS, and when I have tested this it has always been the case that all the headers are rendered before the cells.
> 
> However, I can't actually find anything in the docs that tells me that this will always be the case. I agree that it would feel better if we could find something that tells us that we can rely on this being the case. I will look around more to see if I can find some resources

@nn-davsko I found this resource explaining how the reconciliation algorithm works under the hood in React: https://indepth.dev/inside-fiber-in-depth-overview-of-the-new-reconciliation-algorithm-in-react/#render-phase

It's a bit of a hefty but the part about render phase explains it quite well and it shows the order of which components will be rendered when React traverses the component tree.

tmp2

So it seems like the assumption about rendering order of components is correct and should work as we expect. That being said, this might still be a bad idea. Considering that it is an implementation that relies on an undocumented implementation detail of React, it can be considered quite risky. For example, I don't know how/if this would change with the introduction of Concurrent Mode when that comes.

So I am totally open to scrap this idea completely, if we feel that it is too risky/unstable. What do you think?

toblu:
After discussions within the team, we decided to not go for this implementation since it feels a bit risky and we did not feel that we can 100% guarantee that this solution will work in all cases. Therefore we decided to keep the new approach of applying flex props on each cell, rather than on just the headers

PR #746, opened 2020-10-28
feat: 🎸 added on mouse leave to input text

LifeofMichal:
Added an onMouseLeave focus event to the input text for a special case I have in mortgage. 

Suppose it's a good discussion point if we find a need to add this to other inputs/elements in ui-kit? Or?


victoriatomzik (approved):

victoriatomzik:
good shit

edenvin (approved):

PR #747, opened 2020-10-28
Feature/new link prop

KitsuneKelso:
This change adds the more descriptive `fullServerRedirect` prop intended for use in favour of the now deprecated `cms` prop.


smithsamyan (approved):

crewneck (approved):

src/Molecules/Button/Button.tsx
```diff
@@ -116,6 +117,12 @@ export const Button: ButtonComponent = React.forwardRef<
     },
   );
 
+  if (cms) {
+    assert(false, 'Link: the prop cms is deprecated, please use fullServerRedirect instead.', {
```
KitsuneKelso, line 121: I copy pasted the assert a bit too quickly, this should be `Button` rather than `Link`. 🙂

nn-davsko (approved):

PR #754, opened 2020-11-10
Feat/showmorebutton

victoriatomzik:


src/Molecules/ShowMoreButton/ShowMoreButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  loading = false,
+  align = 'center',
+  showMoreText,
+  showLessText,
+}) => {
+  return (
+    <Flexbox container justifyContent={align === 'left' ? 'flex-start' : 'center'}>
+      <Flexbox item>
+        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
+        <Link onClick={!disabled ? onClick : () => {}} disabled={disabled} aria-expanded={expanded}>
```
kmelkon, line 20: Is this a good idea from an accessibility POV? Can't we have a button that looks like a link? 🤔

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+          {/* eslint-disable-next-line no-nested-ternary */}
```
kmelkon, line 21: I personally think that nested ternaries are very hard to read.
One way to solve this would be to move this into a `useCallback` and using regular if statements to return the correct component

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+          {!expanded ? (
```
kmelkon, line 22: This is a nitpick but I think that having a positive if condition is easier to parse than having a `!` negative condition.

src/Molecules/ShowMoreButton/index.ts
```diff
@@ -0,0 +1,3 @@
+import { ShowMoreButton } from './ShowMoreButton';
+
+export default ShowMoreButton;
```
kmelkon, line 3: IIRC I think you might also need to export this from the main `index.ts`

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
victoriatomzik: @mjmonline what's your take on this one?

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
LifeofMichal: nested ternaries FTW

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
LifeofMichal: I think it depends on context. The click is supposed to be the primary "purpose" of that link ( I assume), so to me (personally) it's easier if it's the primary result of a ternary.

A.k.a. The not part should be the if primary function can't be run, no?

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
LifeofMichal: Personally, I prefer to use neutral buttons with a 'to' attribute. Even if you have to redirect, we can set something like  [this] (https://scm.prod.nordnet.se/projects/DND/repos/ui-mutations/browse/src/page-mortgage-application/ConfirmationStep/ConfirmationStep.tsx), line 134

victoriatomzik: The only thing is - buttons do not take other colors than black/red/blue, and in disabled case I'm gonna need a grey one

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
victoriatomzik: @LifeofMichal yes and yes

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
mjmonline: @kmelkon good and correct thinking 👍. In this case though the`<Link>` component without the `to` prop renders a button looking like a link, so it's already as you want it 🙂
.
@LifeofMichal I would use the `<Link>` component here as well because I get all the styling and it will render a `<button>`

src/Molecules/ShowMoreButton/ShowMoreButton.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+      onClick={() => {}}
+      expanded={false}
+      loading
+      showMoreText="Show more"
+      showLessText="Show less"
+    />
+  );
+};
+
+storiesOf('Guidance / Show More Button', module)
```
mjmonline, line 83: You mean Molecules/...

src/Molecules/ShowMoreButton/ShowMoreButton.types.ts
```diff
@@ -0,0 +1,9 @@
+export type ShowMoreButtonComponent = {
+  onClick?: (val: any) => void;
```
mjmonline, line 2: onClick?: (expanded: boolean) => void;

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+            loading ? (
+              <Spinner
+                id="show-more-spinner"
```
mjmonline, line 25: you might have more than one of this component on a page

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+                color={(t) => (disabled ? t.color.disabledText : t.color.cta)}
+              />
+            ) : (
+              <>
+                {showMoreText}&nbsp;
```
mjmonline, line 30: not a fan of this way of handling spacing, but you decide what you want to do

src/Molecules/ShowMoreButton/ShowMoreButton.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+  .add('Default', () => <MockedShowMoreButton />)
```
mjmonline, line 84: No need to have a mocked component, this is useful in mutations not so much here

src/Molecules/ShowMoreButton/ShowMoreButton.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+          Ex fugit nulla veritatis! Lorem ipsum dolor sit amet, consectetur adipisicing elit.
+          Consectetur distinctio dolore iure nihil quam quis vel. Aut autem consequatur ex impedit
+          magni molestias obcaecati omnis possimus! Ex fugit nulla veritatis! Lorem ipsum dolor sit
+          amet, consectetur adipisicing elit.
+        </Typography>
+      )}
+      <ShowMoreButton
+        onClick={() => setShowMoreA(!showMoreA)}
+        expanded={showMoreA}
+        align="left"
```
mjmonline, line 29: so the default story should be the component with it's default values. Then you can create another story where you change `align`

src/Molecules/ShowMoreButton/ShowMoreButton.stories.tsx
victoriatomzik: I do

src/Molecules/ShowMoreButton/ShowMoreButton.tsx
victoriatomzik: how would you solve this one then? uuid?

src/Molecules/ShowMoreButton/ShowMoreButton.types.ts
victoriatomzik: React.MouseEvent rather

anneback (approved):

mjmonline (approved):

kmelkon (approved):

PR #757, opened 2020-11-11
Feat/date picker with range

fredriklu:
This is the first iteration to avoid one giant PR in the end. This also contains bug fixes for regular date picker that was encountered when implementing the range.

This contains small accessibility updates as well, but these have not been fully focused and will be revised in another PR.


src/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -67,39 +69,74 @@ export const DatePicker = (React.forwardRef<HTMLDivElement, Props>((props, ref)
...
+        : `${format(startDate, dateFormat, options)} - ${format(endDate, dateFormat, options)}`;
+
+      setInputValue(rangeDateString);
+
+      if (onChange) onChange(startDate, endDate);
+    },
+    [selectedDate, dateFormat, options, onChange, selectedEndDate],
+  );
+
+  const handleRegularDateClick = useCallback(
```
nn-davsko, line 123: is there any way the naming of handleRegularDateClick & handleOnDateClick could be more descriptive?

src/Molecules/DatePicker/DatePickerRange.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { RANGE_DATE_PICKER } from './shared/constants';
+
+export default {
+  title: 'Molecules / DatePicker / DatePicker with range',
+  parameters: {
+    component: DatePicker,
+  },
+};
+
+export const Default = () => (
```
nn-davsko, line 13: Are these all the stories that you need to clearly display different states of the range date picker? I'm not sure myself but I just want to point out that according to our DoD we should have stories describing all different states that a component could have (to a reasonable degree of course it's up to you to decide).

nn-davsko (approved):
Looks really good but a fairly large PR I might have missed something.

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -1,139 +1,95 @@
...
-import { getCalendar, getLocale, newDate } from '../shared/dateUtils';
-import { CalendarDayProps, Props } from './Calendar.types';
-
-const StyledWeekDay = styled(Box)`
+import { getCalendar, getLocale } from '../shared/dateUtils';
+import { Props } from './Calendar.types';
+import { FIRST_DAY, LAST_DAY, NUMBER_OF_VISIBLE_DAYS, NUMBER_OF_VISIBLE_WEEKS } from './constants';
+import { CalendarDay } from './CalendarDay';
+
+const calendarDayRefs = [...Array(NUMBER_OF_VISIBLE_WEEKS)].map(() => {
```
toblu, line 11: I'm not sure if there could be potential consequences of using `createRef` directly in the file outside of the component, since `createRef` is typically only used in class components. I'm a little unsure how this would behave with multiple datepickers since the ref object might not be tied to a specific component instance.

I would suggest you create it using `useRef` instead inside the component. You can create a ref that can store an array of elements. Something like this:
```
const calendarDayRefs = useRef(Array(NUMBER_OF_VISIBLE_WEEKS).fill(Array(NUMBER_OF_VISIBLE_DAYS).fill(null)))
```

src/Molecules/DatePicker/Calendar/CalendarDay.tsx
```diff
@@ -0,0 +1,121 @@
...
+        ref={ref}
+        className={className}
+        $disabled={disabled || (typeof enabled === 'boolean' && !enabled)}
+        $selected={selected}
+        $withinRange={withinRange}
+        $isToday={isToday(date)}
+        $edgeDay={edgeDay}
+        onClick={handleOnClick}
+        onKeyDown={onKeyDown}
+        onFocus={() => onFocus()}
```
almightybruce, line 111: not necessary to curry this :)

src/Molecules/DatePicker/Calendar/CalendarDay.tsx
```diff
@@ -0,0 +1,121 @@
...
+      onKeyDown,
+      sameMonth = true,
+      selected,
+      withinRange = false,
+      edgeDay,
+    },
+    ref,
+  ) => {
+    const textColor = (() => {
+      if (disabled || (typeof enabled === 'boolean' && !enabled)) return 'label';
```
almightybruce, line 84: this condition could be made into a const, since its used on 104 as well :)

src/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -47,6 +47,8 @@ export const DatePicker = (React.forwardRef<HTMLDivElement, Props>((props, ref)
     id,
     open: openProp = false,
     selectedDate: selectedDateProp,
+    selectedEndDate: selectedEndDateProp,
```
almightybruce, line 50: is this selectedStartDate when doing a range? if so, it might be worth to document that

src/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -67,39 +69,74 @@ export const DatePicker = (React.forwardRef<HTMLDivElement, Props>((props, ref)
...
+      setInputValue(rangeDateString);
+
+      if (onChange) onChange(startDate, endDate);
+    },
+    [selectedDate, dateFormat, options, onChange, selectedEndDate],
+  );
+
+  const handleRegularDateClick = useCallback(
+    (date: Date) => {
+      setInputValue(format(date, dateFormat, options));
```
almightybruce, line 125: does this still handle controlled behaviour?

src/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -118,31 +155,73 @@ export const DatePicker = (React.forwardRef<HTMLDivElement, Props>((props, ref)
...
-      if (!isMatch(value, dateFormat, opts)) {
-        return;
-      }
+        if (parsedStartDate && parsedEndDate)
+          return parsedStartDate < parsedEndDate
+            ? [parsedStartDate, parsedEndDate]
+            : [parsedEndDate, parsedStartDate];
+
+        return [parsedStartDate, parsedEndDate];
+      })();
```
almightybruce, line 171: maybe add a comment that this is self invoked and why

src/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -67,39 +69,74 @@ export const DatePicker = (React.forwardRef<HTMLDivElement, Props>((props, ref)
...
-      }
+        const swapDate = !selectedEndDate && !isAfter(date, selectedDate);
+        const moveSelectedDate =
+          selectedEndDate &&
+          isSameDay(selectedDate, closestTo(date, [selectedEndDate, selectedDate]));
+
+        if (swapDate) return [date, selectedDate];
+        if (moveSelectedDate) return [date, selectedEndDate];
+        return [selectedDate, date];
+      })();
```
almightybruce, line 107: maybe add a comment that this is self invoked and why

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -145,27 +101,49 @@ const Calendar: React.FC<Props> = ({
...
-              hover={hoverDate && isSameDay(hoverDate, d)}
-            />
-          ))}
+          {week.map((d, dayIndex) => {
+            const edgeDay = (() => {
+              if (weekIndex === 0 && dayIndex === 0) return FIRST_DAY;
+              if (weekIndex === calendar.dates.length - 1 && dayIndex === week.length - 1)
+                return LAST_DAY;
+              return null;
+            })();
```
almightybruce, line 118: maybe add a comment that this is self invoked and why

src/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -67,39 +69,74 @@ export const DatePicker = (React.forwardRef<HTMLDivElement, Props>((props, ref)
...
 
+      setSelectedDate(startDate);
+      setSelectedEndDate(endDate);
+
+      const rangeDateString = !endDate
+        ? `${format(startDate, dateFormat, options)} -`
+        : `${format(startDate, dateFormat, options)} - ${format(endDate, dateFormat, options)}`;
```
gustav-pettersson-nordnet, line 114: Might be nice to add "-" to a constant.

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -145,27 +101,49 @@ const Calendar: React.FC<Props> = ({
...
+                disabled={disableDate && disableDate(d)}
+                onClick={() => onClick(d)}
+                onKeyDown={handleKeyPress}
+                selected={
+                  (selectedDate && isSameDay(selectedDate, d)) ||
+                  (selectedEndDate && isSameDay(selectedEndDate, d))
+                }
+                sameMonth={isSameMonth(viewedDate, d)}
+                locale={localeObj}
+                withinRange={
```
almightybruce, line 139: without knowing what this does, is there a better name for it? like `dateIsWithinCurrentRange`

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -1,139 +1,95 @@
...
-  const enter = useKeyPress('Enter');
-  const [hoverDate, setHoverDate] = useState<Date>(newDate(selectedDate || viewedDate));
+  const [[focusedWeek, focusedDay], setFocused] = useState<[number | null, number | null]>([
+    null,
+    null,
+  ]);
+  const focusedDateObjRef = useRef<Date | null>(null);
+
+  const handleKeyPress = (event: React.KeyboardEvent) => {
+    if (R.isNil(focusedWeek) || R.isNil(focusedDay)) {
```
toblu, line 41: Perhaps it would be a good idea to call `event.stopPropagation()` here? Since you don't want these events to "bubble" up to parent elements, right?

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -1,139 +1,95 @@
...
+      setFocused([0, 0]);
+    } else {
+      switch (event.key) {
+        case 'ArrowLeft':
+          if (focusedDay > 0) {
+            setFocused([focusedWeek, focusedDay - 1]);
+          } else if (focusedWeek !== 0) {
```
toblu, line 48: Any reason that you're using `focusedWeek !== 0` here and not `focusedWeek > 0`? `focusedWeek` should never be less than zero, right?

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -145,27 +101,49 @@ const Calendar: React.FC<Props> = ({
...
-              enabled={enableDate && enableDate(d)}
-              disabled={disableDate && disableDate(d)}
-              onClick={() => onClick(d)}
-              selected={selectedDate && isSameDay(selectedDate, d)}
-              sameMonth={isSameMonth(viewedDate, d)}
-              locale={localeObj}
-              hover={hoverDate && isSameDay(hoverDate, d)}
-            />
-          ))}
+          {week.map((d, dayIndex) => {
```
toblu, line 112: Perhaps rename `d` to `day` here. `d` could be anything

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -145,27 +101,49 @@ const Calendar: React.FC<Props> = ({
...
+                  (selectedDate && isSameDay(selectedDate, d)) ||
+                  (selectedEndDate && isSameDay(selectedEndDate, d))
+                }
+                sameMonth={isSameMonth(viewedDate, d)}
+                locale={localeObj}
+                withinRange={
+                  selectedEndDate &&
+                  isWithinInterval(d, { start: selectedDate, end: selectedEndDate })
+                }
+                edgeDay={edgeDay}
```
toblu, line 143: Perhaps it would be easier to split this into two props: `isFirstDay` & `isLastDay` or something like that? Not so important, so feel free to leave it as it is if you prefer that. Just bringing it up as something you could consider :)

src/Molecules/DatePicker/Calendar/CalendarDay.tsx
toblu, line 84: is the `typeof` check necessary here? in what case is `enabled` not a boolean?

src/Molecules/DatePicker/Calendar/Calendar.tsx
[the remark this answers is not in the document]
robinandersson: The aren't actually the same, spreading like we do yields an array with 7 undefined entries, not spreading the array yields an array with 7 "empty" entries... Don't ask me why, but it does change the execution and we couldn't get it to work without spreading the arrays.

nn-davsko (changes requested):
We need to fix A11y or at least have another look at it. It seems voice over does not work when tab-selecting new dates in the calendar view.

src/Molecules/DatePicker/DatePicker.tsx
fredriklu: Do you mean `selectedDate`? If so: Yes. It has kept that naming since `selectedStartDate` would be confusing for the regular DatePicker.

Do you propose leaving a comment?

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -1,139 +1,95 @@
...
+
+        case 'ArrowDown':
+          if (focusedWeek < NUMBER_OF_VISIBLE_WEEKS - 1) {
+            setFocused([focusedWeek + 1, focusedDay]);
+          }
+          break;
+
+        case ' ': // Spacebar
+        case 'Enter':
+          if (!R.isNil(focusedDateObjRef.current)) onClick(new Date(focusedDateObjRef.current));
```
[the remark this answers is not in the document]
robinandersson: If it's the second part you're referring to, i.e. what we pass to the onClick handler then yes, it will be set. But we still need to create a new Date object to create a new reference (onClick handler might mutate focus object, thus opening up the potential for a whole lot of worms).

fredriklu:
> We need to fix A11y or at least have another look at it. It seems voice over does not work when tab-selecting new dates in the calendar view.

@nn-davsko This is known, see the first comment/description of the PR. Since this PR is already quite large, and/but still improves the existing regular DatePicker which seems to lack A11y features as well, we decided to create the PR anyway.

src/Molecules/DatePicker/DatePicker.tsx
fredriklu: This seems excessive. This is a quite common pattern and adding an explanation of IIFE and why would be quite a meaty comment. 

An alternative would be refactoring it to a proper function, but we feel that would add unnecessary complexity.

src/Molecules/DatePicker/DatePicker.tsx
fredriklu: See reply on first instance.

src/Molecules/DatePicker/Calendar/Calendar.tsx
fredriklu: See reply on first instance.

src/Molecules/DatePicker/DatePicker.tsx
fredriklu: Do you mean create e.g. const INPUT_SEPARATOR = '-'?

src/Molecules/DatePicker/Calendar/Calendar.tsx
fredriklu: Since it only handles styling (the grey background and edge of month fade), and there can only be one range, therefore this seems a bit excessive, especially since the code below also indicates what it controls.

nn-davsko:
> > We need to fix A11y or at least have another look at it. It seems voice over does not work when tab-selecting new dates in the calendar view.
> 
> @nn-davsko This is known, see the first comment/description of the PR. Since this PR is already quite large, and/but still improves the existing regular DatePicker which seems to lack A11y features as well, we decided to create the PR anyway.

Ok sounds fair, I missed that in the comments.

nn-davsko (approved):

src/Molecules/DatePicker/DatePicker.tsx
almightybruce: Yes

src/Molecules/DatePicker/Calendar/Calendar.tsx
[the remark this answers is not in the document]
almightybruce: So i think you should explain that in a comment so nobody refactors it away

src/Molecules/DatePicker/Calendar/Calendar.tsx
almightybruce: It doesn't hurt to be clear. I think you should at least put an `is` infront

src/Molecules/DatePicker/DatePicker.tsx
almightybruce: Not saying you should explain IIFE's and their existence. Just saying the `()` is easily missed and since there's no obvious reason to do an IIFE, a heads up would have been nice.

src/Molecules/DatePicker/Calendar/Calendar.tsx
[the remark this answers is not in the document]
toblu: Just a comment regarding why it was necessary to use the spread operator here. The `Array()` function creates an array of the specified length, however it does not give the elements any value (`undefined` is a value). Array elements without a value are not enumerable and that is why `.map()` has no effect on an array with only `empty` elements. When you use the spread operator, all the elements in the newly created array are initialized with the value `undefined` and that is the reason why you are able to use `.map()` on the new array

fredriklu: .fill did not create unique ref:s, but useRef is now implemented!

src/Molecules/DatePicker/Calendar/Calendar.tsx
```diff
@@ -1,139 +1,97 @@
...
-  const [hoverDate, setHoverDate] = useState<Date>(newDate(selectedDate || viewedDate));
+  const [[focusedWeek, focusedDay], setFocused] = useState<[number | null, number | null]>([
+    null,
+    null,
+  ]);
+  const focusedDateObjRef = useRef<Date | null>(null);
+
+  const calendarDayRefs = useRef(
+    [...Array(NUMBER_OF_VISIBLE_WEEKS)].map(() => {
+      return [...Array(NUMBER_OF_VISIBLE_DAYS)].reduce((acc) => {
```
toblu, line 36: Doesn't this reduce function behave just like a regular .map here? In that case I would say it would be easier to understand this line if you only used .map.

`[...Array(NUMBER_OF_VISIBLE_DAYS)].map(React.createRef)`

src/Molecules/DatePicker/Calendar/Calendar.tsx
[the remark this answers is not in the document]
robinandersson: Ooh, nice explanation @toblu! Did not know this :)

almightybruce (approved):
stellar work!

src/Molecules/DatePicker/DatePicker.tsx
fredriklu: After further discussion, we decided to leave it as is.

src/Molecules/DatePicker/DatePicker.tsx
fredriklu: After further discussion, we don't think requires special treatment more than any other pattern. We have not made these kind of comments for e.g. curried functions previously.

src/Molecules/DatePicker/DatePicker.tsx
gustav-pettersson-nordnet: Yes :) Mostly to see that is not a subtraction :) but the char "-"

toblu (approved):
Great job guys! ⭐

robinandersson:
Thanks! 🙂

frozzare (approved):

src/Molecules/DatePicker/DatePicker.tsx
fredriklu: This seems kind of extensive for just one character, especially since it's within a template string that's named rangeDateString

PR #759, opened 2020-11-13
Feat/showmore

victoriatomzik:


src/Organisms/ShowMore/ShowMore.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  const onShowMoreClick = useCallback(
+    (e) => {
+      onShowMore(e);
+      setShowMoreClicked(true);
+    },
+    [onShowMore],
+  );
+
+  // @ts-ignore
```
kmelkon, line 43: this is not a good sign usually :D what's it complaining about?

src/Organisms/ShowMore/useOverflow.ts
```diff
@@ -0,0 +1,27 @@
+import { useState, useLayoutEffect } from 'react';
+import { throttle } from 'lodash';
```
kmelkon, line 2: I'm pretty sure there's a way to throttle without adding a new dependancy.

kmelkon (changes requested):

kmelkon:
Looks good! just left a couple of comments

src/Organisms/ShowMore/ShowMore.tsx
victoriatomzik: dude... idek
TS2322: Type 'ReactNode' is not assignable to type 'never'. Type 'undefined' is not assignable to type 'never'.

src/Organisms/ShowMore/useOverflow.ts
victoriatomzik: ;_;

src/Organisms/ShowMore/ShowMore.tsx
kmelkon: hmm we can look into it later but usually TS throws an error cause there's an underlying issue

alvelid (approved):

jesperwiner (approved):

PR #760, opened 2020-11-13
fix: 🐛 remove left padding for quiet number input

kmelkon:


victoriatomzik (approved):

victoriatomzik:
gg wp

jesperwiner (approved):

src/Molecules/Input/Number/Number.tsx
```diff
@@ -139,7 +135,7 @@ const Input = styled(NormalizedElements.Input).attrs(() => ({ type: 'text' }))<P
          font-size: 28px; 
          font-weight: bold;
          &:focus {
-           padding-left: ${p.theme.spacing.unit(p.leftAddon ? 8 : 2)}px;
+           padding-left: ${p.theme.spacing.unit(p.leftAddon ? 3 : 2)}px;
```
fredriklu, line 138: Is there any reason why `leftAddon` is still left here? Perhaps this constant could be named into something more appropriate?

src/Molecules/Input/Number/Number.tsx
```diff
@@ -76,8 +76,7 @@ const AddonBox = styled(Flexbox)<{ position?: 'left' | 'right' }>`
   top: 0;
   height: 100%;
   z-index: 3;
-  ${(p) => (p.position === 'left' ? `left: ${p.theme.spacing.unit(2)}px;` : '')}
-  ${(p) => (p.position === 'right' ? `right: ${p.theme.spacing.unit(2)}px;` : '')}
+  ${(p) => (p.position === 'right' ? `right: ${p.theme.spacing.unit(0)}px;` : '')}
```
LifeofMichal, line 79: What's this about? Do you need to take away this right for the padding to work?

victoriatomzik:
@kmelkon Should we do something about it?

kmelkon:
> @kmelkon Should we do something about it?

completely forgot about this, I'll check it now

src/Molecules/Input/Number/Number.tsx
kmelkon: It's so the right addOn is aligned with the bottom border when not focused. Forgot to mention that in the PR description.
image

src/Molecules/Input/Number/Number.tsx
kmelkon: Will fix, you're correct, it's not needed when using `quiet` anymore

Vikingjonsson (approved):

PR #765, opened 2020-11-18
FlexTable: Remove RenderForSizez

toblu:
Remove `RenderForSizes` component in FlexTable and replace it with a CSS solution that generates all the styles for different screen sizes. What this means is that instead of rendering separate elements for each screen size, we render only one element no matter screen size and use CSS to style properly for each screen size using css media queries. This leads to a significant reduction in number of elements that have to be rendered (especially on the server), and thereby increases rendering performance. 

This also fixes a hydration bug between the server and the client, since the table previously did not render the same way on the server as the first render in the client. This can lead to unexpected issues since React can't patch the differences and expects the rendered content to be identical

> React expects that the rendered content is identical between the server and the client.

Read more: https://reactjs.org/docs/react-dom.html#hydrate

These changes are mostly internal changes in the FlexTable, but it has a few implications on the external API. These are:
- Different `expandItems` / `expandChildren` can no longer be passed as `MediaRelatedProps` (`sm`, `md`, `lg`, `xl`) on the Row. Instead if you need to configure different `expandItems` for different screen sizes, you can do it by setting the `hidden` prop directly on an expand item for a specific size (see docs / stories for example)
- You can no longer use `fontSize` render prop for cell content, the Cell's TextWrapper component automatically gets the `fontSize` from the table


nn-davsko (approved):
Great job guys @toblu @mhogler @gustav-pettersson-nordnet !!!

almightybruce (approved):
maxresdefault

PR #780, opened 2020-11-30
Feat/date range picker double input

almightybruce:
This PR includes basic functionality for the double date picker. It also includes a restructuring of the folder structure and a unification of the date/range/double pickers.

Some tests for the double picker are there but more will be written.


src/Molecules/DatePicker/DatePicker.tsx
```diff
@@ -1,386 +1,20 @@
...
-      isUndefined(disableDate),
-      `DatePicker: "disableDate" cannot be used at the same time as "enableDate".`,
-    );
+import React from 'react';
+import DoubleDatePicker from './Double';
+import DatePicker from './Single';
+import { DOUBLE_DATE_PICKER } from './shared/constants';
+import { SingleDatePickerProps, DoubleDatePickerProps } from './DatePicker.types';
+
+export function isDoubleDatePicker(
```
toblu, line 7: nice type guard! :)

src/Molecules/DatePicker/Double/Calendar/DoubleCalendar.tsx
```diff
@@ -0,0 +1,251 @@
...
+  text-align: center;
+  margin-bottom: ${({ theme }) => theme.spacing.unit(2)}px;
+  margin-top: ${({ theme }) => theme.spacing.unit(3)}px;
+`;
+
+const HiddenDate = styled.div<{ $withGutter?: boolean }>`
+  min-width: ${({ theme }) => theme.spacing.unit(10) + 2}px;
+  min-height: ${({ theme }) => theme.spacing.unit(10) + 2}px;
+  margin: ${({ theme }) => theme.spacing.unit(0.5)}px 0;
+  ${({ $withGutter = false }) => ($withGutter ? 'margin-right: 34px;' : '')}
```
toblu, line 36: Nit-picking here, but is it necessary to set `$withGutter` to default value false here? `undefined` is a also a falsy value so it shouldn't make any difference right? I'm thinking that skipping the default value might be better for readability (same on line 40)

src/Molecules/DatePicker/Double/Calendar/DoubleCalendar.tsx
```diff
@@ -0,0 +1,251 @@
...
+  controlledFocus,
+}) => {
+  const [[focusedWeek, focusedDay], setFocused] = useState<[number | null, number | null]>([
+    null,
+    null,
+  ]);
+  const focusedDateObjRef = useRef<Date | null>(null);
+
+  const dateIsDisabled = (dateToCheck: Date | null) => {
+    const isEnabled = dateToCheck && enableDate && enableDate(dateToCheck);
```
toblu, line 60: This approach is fine, but if you wanted to make it a little shorter you could also use optional chaining here.
`const isEnabled = dateToCheck && enableDate?.(dateToCheck);`

src/Molecules/DatePicker/shared/components/CalendarDay/CalendarDay.types.ts
```diff
@@ -0,0 +1,16 @@
+export type CalendarDayProps = {
+  className?: string;
+  date: Date;
+  disabled?: boolean;
+  enabled?: boolean;
+  locale: any;
```
robinandersson, line 6: any way to avoid any type ?

src/Molecules/DatePicker/DatePickerUnified.stories.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { action } from '@storybook/addon-actions';
+import DatePicker from './DatePicker';
+
+export default {
+  title: 'Molecules / DatePicker / Unified',
```
fredriklu, line 6: I don't think this is how we have done it previously. If you'd wan't to see every DatePicker side by side, I'd suggest that `Molecules / DatePicker` contain a story "All" where they're all rendered within the same, and then each variant have their own folder

src/Molecules/DatePicker/Double/DoubleDatePicker.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export const DisallowSingleDayRange = () => (
+  <DoubleDatePicker
+    id="input-id"
+    labelFrom="Label"
+    labelTo=""
+    onChange={action('Range date')}
+    disallowSingleDayRange
+    variant="DOUBLE"
```
nn-davsko, line 102: why does DoubleDatePicker have a variant that says DOUBLE?

src/Molecules/DatePicker/Double/DoubleDatePicker.types.ts
```diff
@@ -0,0 +1,23 @@
...
+  disableDate?: (date: Date) => boolean;
+  disabled?: boolean;
+  enableDate?: (date: Date) => boolean;
+  id: string;
+  selectedStartDate?: Date;
+  selectedEndDate?: Date;
+  inputValueStart?: string;
+  inputValueEnd?: string;
+  inputSize?: 's';
+  variant: 'DOUBLE';
```
robinandersson, line 17: Is this variant prop needed?

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
```diff
@@ -0,0 +1,353 @@
...
+    id,
+    open: openProp = false,
+    selectedStartDate: selectedStartDateProp,
+    selectedEndDate: selectedEndDateProp,
+    inputValueStart: inputValueStartProp,
+    inputValueEnd: inputValueEndProp,
+    width = 78,
+    yearSelectLength,
+    inputSize,
+    disallowSingleDayRange = false,
```
nn-davsko, line 65: There is no "variant" prop here is it used?

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
```diff
@@ -0,0 +1,353 @@
...
+const StyledInputTextLeft = styled(Input.Text)`
+  z-index: 1;
+  padding-right: ${({ theme }) => theme.spacing.unit(INPUT_SPACING / 2)}px;
+`;
+
+const StyledDropdownBubble = styled(DropdownBubble)`
+  width: ${({ theme }) => theme.spacing.unit(156 + INPUT_SPACING)}px;
+  z-index: ${({ theme }) => theme.zIndex.overlay};
+  box-sizing: border-box;
+  top: -10px;
```
gustav-pettersson-nordnet, line 35: maybe using theme.spacing.unit(2.5) here? But it's not a whole number, so not sure.

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
```diff
@@ -0,0 +1,353 @@
...
+    disabled,
+    disableDate,
+    enableDate,
+    id,
+    open: openProp = false,
+    selectedStartDate: selectedStartDateProp,
+    selectedEndDate: selectedEndDateProp,
+    inputValueStart: inputValueStartProp,
+    inputValueEnd: inputValueEndProp,
+    width = 78,
```
gustav-pettersson-nordnet, line 62: would be nice to use a constant here or use of unit.

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
```diff
@@ -0,0 +1,353 @@
...
+    yearSelectLength,
+    inputSize,
+    disallowSingleDayRange = false,
+  } = props;
+
+  assert(Boolean(props.id), `DatePicker: "id" is required.`);
+
+  if (disableDate) {
```
toblu, line 70: Do you actually need both assertions here? If the first one fails, then the second one will always fail, and vice versa if the first one passes then the second one will always pass. Or is there a specific reason that you want to have to separate ones? :)

src/Molecules/DatePicker/Double/DoubleDatePicker.types.ts
nn-davsko, line 17: I keep going on about this :)

src/Molecules/DatePicker/shared/components/CalendarDay/CalendarDay.tsx
```diff
@@ -12,11 +12,13 @@ const StyledCalendarDay = styled(Box)<{
   $isSameMonth?: boolean;
   $isFirstDay?: boolean;
   $isLastDay?: boolean;
+  $withGutter?: boolean;
 }>`
   background: ${({ theme }) => theme.color.backgroundInput};
   min-width: ${({ theme }) => theme.spacing.unit(10) + 2}px;
   min-height: ${({ theme }) => theme.spacing.unit(10) + 2}px;
   margin: ${({ theme }) => theme.spacing.unit(0.5)}px 0;
+  ${({ $withGutter = false }) => ($withGutter ? 'margin-right: 34px;' : '')}
```
gustav-pettersson-nordnet, line 21: would be nice to a constant or use of unit here.

gustav-pettersson-nordnet (approved):

src/Molecules/DatePicker/Double/Header/DoubleHeader.tsx
```diff
@@ -0,0 +1,110 @@
...
+  width: ${({ theme }) => theme.spacing.unit(76)}px;
+`;
+
+const ChevronContainer = styled(Box)<{ $align: 'left' | 'right' }>`
+  position: absolute;
+  ${({ $align, theme }) => `${$align}: ${theme.spacing.unit(2)}px;`}
+`;
+
+const DoubleHeader: React.FC<Props> = ({
+  ariaLabelPrevious = 'Previous month {date}',
```
nn-davsko, line 21: I don't know if this needs to be "solved" but should these be translated?

src/Molecules/DatePicker/Single/DatePicker.tsx
```diff
@@ -0,0 +1,388 @@
...
+import { RANGE_DATE_PICKER, REGULAR_DATE_PICKER } from '../shared/constants';
+
+const StyledInputText = styled(Input.Text)`
+  z-index: 1;
+`;
+
+const StyledDropdownBubble = styled(DropdownBubble)`
+  max-width: ${({ theme }) => theme.spacing.unit(78)}px;
+  z-index: ${({ theme }) => theme.zIndex.overlay};
+  top: -10px;
```
nn-davsko, line 33: units?

fredriklu:
To improve clarity regarding the file structure, perhaps the files should be renamed for single DatePicker, so it's RegularDatePicker, RangeDatePicker and DoubleDatePicker, instead of DatePicker, DatePickerRange and DoubleDatePicker?

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
```diff
@@ -0,0 +1,353 @@
...
+        autoComplete="off"
+      />
+      {open ? (
+        <StyledDropdownBubbleWrapper>
+          <StyledDropdownBubble>{datepicker}</StyledDropdownBubble>
+        </StyledDropdownBubbleWrapper>
+      ) : null}
+    </div>
+  );
+}) as any) as React.FC<Props> & {};
```
toblu, line 351: Why do you have this type casting first to `any` and then back to `React.FC`? 🤔  Is it actually necessary?

robinandersson (approved):

nn-davsko (approved):

src/Molecules/DatePicker/Single/DatePicker.tsx
```diff
@@ -0,0 +1,388 @@
...
+    inputValue: inputValueProp,
+    variant = REGULAR_DATE_PICKER,
+    width = 78,
+    yearSelectLength,
+    inputSize,
+  } = props;
+
+  assert(Boolean(props.id), `DatePicker: "id" is required.`);
+
+  if (disableDate) {
```
toblu, line 66: Same comment here about the necessity of double assertions

src/Molecules/DatePicker/Single/DatePicker.tsx
```diff
@@ -0,0 +1,388 @@
...
+        autoComplete="off"
+      />
+      {open ? (
+        <StyledDropdownBubbleWrapper>
+          <StyledDropdownBubble>{datepicker}</StyledDropdownBubble>
+        </StyledDropdownBubbleWrapper>
+      ) : null}
+    </div>
+  );
+}) as any) as React.FC<Props> & {};
```
toblu, line 386: Same question here regarding type casting

toblu (approved):
Great job, looks really solid! ⭐  Left a few comments, but only minor stuff. Great that you added tests for the DoubleDatePicker as well!

src/Molecules/DatePicker/DatePicker.tsx
almightybruce: thanks! :)

src/Molecules/DatePicker/Double/Calendar/DoubleCalendar.tsx
almightybruce: good catch!

src/Molecules/DatePicker/Double/Calendar/DoubleCalendar.tsx
almightybruce: we'll try that, it looks neater

src/Molecules/DatePicker/Double/DoubleDatePicker.stories.tsx
almightybruce: it shouldn't!

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
almightybruce: It was copied from the other datepicker, but we'll look into it :)

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
almightybruce: Can do. Note to self: fix in regular datepicker too

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
almightybruce: Can do. Note to self: fix in regular datepicker too

src/Molecules/DatePicker/shared/components/CalendarDay/CalendarDay.tsx
almightybruce: Can do, but it wont be whole units.

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
almightybruce: Interesting, we'll look at that. It was copied from regular so note to self: fix that too

src/Molecules/DatePicker/shared/components/CalendarDay/CalendarDay.types.ts
almightybruce: Will change to Locale type :)

src/Molecules/DatePicker/Double/DoubleDatePicker.types.ts
almightybruce: Niet!

almightybruce: and we appreciate it!

src/Molecules/DatePicker/Double/Header/DoubleHeader.tsx
almightybruce: Ouf, yes. Note to self: fix in regular as well

almightybruce: Or... I guess? What are the thoughts on using our l10n in ui?

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
mhogler: We decided to not use units for this case. The -10px is used to align the calendar-part with the input fields which is misaligned because of `transform: translateY(${TOP_OFFSET}px);` from the DropDownBubble atom in UI. In DropDownBubble `TOP_OFFSET` is hardcoded to 10 and is not related to units. Rather than using units, we suggest that DropDownBubble atom should be fixed first

src/Molecules/DatePicker/Double/Header/DoubleHeader.tsx
toblu: I don't think you can use the l10n package in this repo unfortunately. I know that there were discussions on how to handle translations in ui, but I don't know what the conclusions were (or if there was any)? In the Pagination component we at least solved it by adding props for the labels so that they can be set from outside. AFAIK we currently have no better way of handling it

PR #800, opened 2020-12-08
Fix/grey quiet placeholder

kmelkon:
Grey out quiet input's placeholder on focus


src/Molecules/Input/Number/Number.tsx
```diff
@@ -18,7 +18,7 @@ const width = css<Pick<Props, 'size'>>`
 `;
 
 const height = css<Pick<Props, 'size'>>`
-  height: ${(p) => (p.size === 's' ? p.theme.spacing.unit(8) : p.theme.spacing.unit(10))}px;
+  height: ${(p) => p.size === 's' && p.theme.spacing.unit(8)}px;
```
ratkosolar, line 21: using `&&` shorthand causes weird issues in styled components because it returns false when the condition is not met... so basically when size is not 's' it will output invalid css like seen in screenshot
image

kmelkon: Ah I didn't know that, should I just use `''` when the condition is not met?

ratkosolar: maybe something like `${(p) => p.size === 's' ? `height: ${p.theme.spacing.unit(8)}` : '' }`
not sure what does this have to do with graying out the placeholder though

victoriatomzik: jaha, do the ? : dance

kmelkon: It didn't adhere to the design for quiet number inputs, but I'll do that and see if it breaks other variants

src/Molecules/Input/Text/Text.tsx
```diff
@@ -65,6 +65,9 @@ export const placeholderNormalizaion = css<Pick<Props, 'variant'>>`
     line-height: inherit;
     opacity: 1;
   }
+  &:focus::placeholder {
+    color: ${(p) => p.variant === 'quiet' && p.theme.color.disabledText};
```
victoriatomzik, line 69: what is this going to be resolved to when the values are !'quiet' and !disabledText?

kmelkon: I assume it's gonna use the label colour that's define a few lines above this line but I should add a `? :` to be safe

victoriatomzik: yeah, either this or you can just add the whole rule in the condition "color" included, so the whole line is kinda on/off, and there's no situation that it returns something like null/undefined

kmelkon: Hmm that sounds nicer, I'll try that

victoriatomzik (approved):
NOICE

src/Molecules/Input/Text/Text.tsx
```diff
@@ -58,13 +58,16 @@ const borderStyles = css<Pick<Props, 'error' | 'success' | 'disabled' | 'variant
...
+export const placeholderNormalizaion = css<Pick<Props, 'variant' | 'disabled'>>`
   &::placeholder {
     color: ${(p) => (p.variant === 'quiet' ? p.theme.color.cta : p.theme.color.label)};
     height: inherit;
     line-height: inherit;
     opacity: 1;
   }
+  ${(p) =>
+    p.variant === 'quiet' ? `&:focus::placeholder { color: ${p.theme.color.disabledText}}` : ''};
+  ${(p) => (p.disabled ? `&:disabled::placeholder { color: ${p.theme.color.disabledText}}` : '')}
```
ratkosolar, line 70: is a combination of `p.disabled` and `&:disabled` necessary?
wouldn't just one be enough or am I wrong?

src/Molecules/Input/Number/Number.tsx
```diff
@@ -17,8 +17,9 @@ const width = css<Pick<Props, 'size'>>`
   width: ${(p) => (p.size === 's' ? p.theme.spacing.unit(8) : p.theme.spacing.unit(10))}px;
 `;
 
-const height = css<Pick<Props, 'size'>>`
+const height = css<Pick<Props, 'variant' | 'size'>>`
   height: ${(p) => (p.size === 's' ? p.theme.spacing.unit(8) : p.theme.spacing.unit(10))}px;
+  height: ${(p) => (p.variant === 'quiet' ? 'auto' : '')};
```
ratkosolar, line 22: but now you have double height property, technically it should work but there is a scenario where this will end up being this
```
height: 40px;
height: ;
```
which is invalid CSS... not sure how styled components will deal with it thought but its safer to do this instead if I got your intention right
```
// its ugly af, I know
const height = css<Pick<Props, "variant" | "size">>`
  ${(p) =>
    p.variant === 'quiet'
      ? 'height: auto'
      : p.size === 's'
        ? `height: ${p.theme.spacing.unit(8)}px`
        : `height: ${p.theme.spacing.unit(10)}px`}
`;

// but you can also do it this way (a bit nicer)
const calculateHeight = (p) => {
  if (p.variant === 'quiet') return 'auto';
  return p.size === 's'
    ? `${p.theme.spacing.unit(8)}px`
    : `${p.theme.spacing.unit(10)}px`;
};

const height = css<Pick<Props, 'variant' | 'size'>>`
  height: ${(p) => calculateHeight(p)};
`;

```

ratkosolar (changes requested):

src/Molecules/Input/Text/Text.tsx
kmelkon: Might not be necessary, I can check. but I know that without the `:disabled` it doesn't work

src/Molecules/Input/Number/Number.tsx
kmelkon: bruh <3 I'll change it

src/Molecules/Input/Text/Text.tsx
ratkosolar: ah okay, was just curious :D

kmelkon: classic

kmelkon: the prop check wasn't needed :D removed it

victoriatomzik (approved):

mjmonline (approved):
LGTM

ratkosolar (approved):

PR #802, opened 2020-12-10
Feat/bottom nav

victoriatomzik:


jesperwiner (approved):

ratkosolar (approved):

mjmonline (commented):
You need to cover more cases with stories. Also this component was named BottomNav in mutations on the page that was using it and thats fine but here it's a global component. It looks like this component is only intended to be used in a guide right? Because you have the submit on last step and cancel button and so on. Also if we are "navigating" we should use links and this case is more like stepping through a guide

src/Molecules/BottomWizardBar/BottomWizardBar.types.ts
```diff
@@ -0,0 +1,20 @@
+export type Props = {
+  onCancel: React.MouseEventHandler;
+  onPrevious: React.MouseEventHandler;
+  onNext: React.MouseEventHandler;
+  onSubmit?: React.MouseEventHandler;
+  /** ^ Switches forward button to submit */
+  hidePreviousButton?: boolean;
+  isLastStep?: boolean;
+  isLoading?: boolean;
+  isEmbedded?: boolean;
```
mjmonline, line 10: you should have /** @default false */ before each of these rows

mjmonline (approved):

PR #807, opened 2020-12-15
Bottomwizardbar 3

victoriatomzik:


src/Molecules/BottomWizardBar/BottomWizardBar.tsx
```diff
@@ -2,16 +2,31 @@ import React from 'react';
 import styled from 'styled-components';
 import { Flexbox, useMedia, Box } from '../..';
 import Button from '../Button';
+import PageWrapper from '../PageWrapper';
 import { Props } from './BottomWizardBar.types';
 
+const tenColumns = `${(10 / 12) * 100}%`; // we want to utilize 10 out of 12 columns
```
ratkosolar, line 8: I think you should use Flexbox component for this instead... also if I remember correctly from the design meeting not always will this component take 10 columns, it might be 12 in some scenarios no?

ratkosolar (changes requested):

src/Molecules/BottomWizardBar/BottomWizardBar.tsx
victoriatomzik: we've discussed it with Patience, this is the wanted component (due to the `getBreakpointStyles` function). it takes 12 until `${(p) => p.theme.media.greaterThan(p.theme.breakpoints.xl)}`, then it's 10 columns to avoid the button groups being too far from one another. ^_^
also, the behaviour of the `PageWrapper` component is consistent with the `PageWrapper` wrapper on the wizard layout, so we don't experience incompatible width jumping.

ratkosolar (approved):

src/Molecules/BottomWizardBar/BottomWizardBar.tsx
```diff
@@ -27,7 +42,7 @@ const BottomWizardBar: React.FC<Props> = ({
   previousText,
   nextText,
   submitText,
-  hidePreviousButton = false,
+  hidePreviousButtonOnMobile = false,
```
mjmonline, line 45: Careful, this is a breaking change

src/Molecules/BottomWizardBar/BottomWizardBar.tsx
```diff
@@ -2,16 +2,31 @@ import React from 'react';
...
+  z-index: ${(p) => p.theme.zIndex.overlay};
+  border-top: 1px solid ${(p) => p.theme.color.divider};
+`;
+
+const StyledBox = styled(Box)`
+  margin: 0 auto;
+  ${(p) => p.theme.media.greaterThan(p.theme.breakpoints.xl)} {
+    width: ${tenColumns};
+  }
 `;
```
mjmonline, line 30: `Box` component without padding or margin attributes like `<Box p={5} />` is just a div, so I would prefer you use `styled.div` here instead or style the `<nav />`

src/Molecules/BottomWizardBar/BottomWizardBar.tsx
victoriatomzik: I know ^-^

src/Molecules/BottomWizardBar/BottomWizardBar.tsx
victoriatomzik: good one, thanks

mjmonline (approved):

PR #810, opened 2020-12-16
feat/move tooltip to popper.js

crewneck:
https://jira.prod.nordnet.se/browse/DND-570  

We want to move away from @reachjs/tooltip since it currently does not support touch events. We decided on adding Popper.js since it has nice handling of Tooltip placement baked in. We did have to handle the logic of opening and closing the tooltip on our own, which was heavily inspired by @reachjs/tooltip source code.  

Visible changes are that you now can open tooltips on mobile (by touching the trigger element) and a new prop called “mode”, which allows the developer to decide if the tooltip should be triggered by hover or by click. Not providing this prop defaults to hover for backwards compatibility.


src/Molecules/Tooltip/Tooltip.tsx
```diff
@@ -1,118 +1,70 @@
...
-  return {
-    left: `${leftCenter(triggerRect, tooltipRect)}px`,
-    top: `${topUnder(triggerRect, tooltipRect)}px`,
-  };
-};
+/** 
+  There are a few features that are important to understand.
+  
+  1. Tooltips don't show up until the user has rested on one, we don't
+     want tooltips popupping up as you move your mouse around the page.
```
victoriatomzik, line 11: popping up*

nn-davsko (approved):

scripts/start-storybook.sh
```diff
@@ -6,4 +6,5 @@ then
     export STORYBOOK_DIRECTORY=$1
 fi
 
-yarn start-storybook -p 6006 -s ./static
\ No newline at end of file
+# --no-dll is required to make popper-react work with storybook because storybook has the same dependency with older version
```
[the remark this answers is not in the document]
ratkosolar: we have considered doing that too but after some thinking it seems like it might cause it to break again in the future if the version we use goes out of sync with the storybook one

src/Molecules/Tooltip/Tooltip.stories.tsx
```diff
@@ -96,7 +107,7 @@ export const zindexWars = () => {
     return (
       <Drawer>
         <Box mb={4}>
-          <Tooltip label={label}>
+          <Tooltip inModal label={label} position="left">
```
ratkosolar, line 110: I think this was added by accident while testing the new tooltip

ratkosolar (changes requested):
image
I just noticed that the build script is broken due to storybook having the same dependency for react-popper but older version.... I remember we did that `--no-dll` in the dev script, maybe we can do the same for build?

victoriatomzik (changes requested):
image

ratkosolar (approved):

victoriatomzik (approved):

PR #814, opened 2020-12-17
Fix/datepicker full width

ratkosolar:


frozzare (approved):

almightybruce (approved):
Looks good but the useMemo seems unnecessary :)

anneback (approved):

PR #816, opened 2020-12-22
Fix/select dropdown popper js

ratkosolar:
We added 2 new things to Input.Select:

1. `placement` prop which can be 'top' or 'bottom' and it defaults to 'bottom'
2. `withPortal` prop which renders the dropdown inside of a Portal using popper.js, really useful for rendering Select dropdowns in a Modal for example.


LifeofMichal (approved):
There are some "missing props", but not in files you were working with, so with that, looks fine-ish.

crewneck (approved):

gustav-pettersson-nordnet (approved):

nn-davsko (approved):
Good job a much needed improvement!

PR #828, opened 2021-01-21
feat: add open and close delay props to tooltip

jesperwiner:


src/Molecules/Tooltip/hooks/useTooltip.ts
```diff
@@ -24,19 +24,20 @@ export const useTooltip = (mode: TooltipProps['mode']) => {
 
   const handleMouseEnter = useCallback(() => {
     if (mode !== 'hover') return;
+    console.log('store', store);
```
Vikingjonsson, line 27: log

kmelkon (approved):

src/Molecules/Tooltip/hooks/useTooltip.ts
```diff
@@ -24,19 +24,20 @@ export const useTooltip = (mode: TooltipProps['mode']) => {
...
     if (!store.isVisible()) {
-      store.startBecomingVisibleTimeout(id);
+      store.startBecomingVisibleTimeout(id, openDelay);
     } else if (store.state === State.LEAVING_VISIBLE) {
       store.setState(State.VISIBLE, id);
     }
-  }, [id, mode]);
+  }, [id, mode, openDelay]);
 
   const handleMouseMove = useCallback(() => {
```
Vikingjonsson, line 36: if this should act the same as handleMouseEnter (maybe enough with one of them) then is probably good to have the same dependencies.

src/Molecules/Tooltip/hooks/useTooltip.ts
jesperwiner: removed

src/Molecules/Tooltip/hooks/useTooltip.ts
jesperwiner: fixed

Vikingjonsson (approved):

PR #841, opened 2021-02-25
Feat/persist sort order for flex table

gustav-pettersson-nordnet:


src/Molecules/FlexTable/shared/FlexTableProvider/FlexTableProvider.types.ts
```diff
@@ -26,6 +31,11 @@ export type FlexTableState = {
    * @default false
    */
   expandable: boolean;
+  /**
+   * Remember last sorted column. The last sorted column will be stored in local storage and be the initial sorted column.
+   * @default false
+   */
+  persistSortingOrder: boolean;
```
gustav-pettersson-nordnet, line 38: persistSortingOrder is part of FlexTableContextProps what I can see but I still get typescript error about It.

src/Molecules/FlexTable/shared/persistedSortOrder.ts
```diff
@@ -0,0 +1,30 @@
+import { SortOrder } from '../Header/HeaderContent/HeaderContent.types';
+import { persistedSortOrderLocalStorageKey } from './constants';
+
+export const tableHasSavedPersistedSortOrder = (tableId: string | undefined) => {
+  if (!tableId) {
+    return false;
+  }
+
+  const sortedSortOrders = localStorage.getItem(persistedSortOrderLocalStorageKey);
```
jesperwiner, line 9: add a try/catch around all access to local storage please

src/Molecules/FlexTable/shared/persistedSortOrder.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (!parsed) {
+    return false;
+  }
+  return !!parsed[tableId];
+};
+
+export const getPersistedSortOrder = (
+  tableId: string | undefined,
+): { columnId: string; sortOrder: SortOrder } | null => {
+  if (tableId && tableHasSavedPersistedSortOrder(tableId)) {
```
jesperwiner, line 24: here too, we have found that some browsers will throw exceptions there that bubble all the way to the oops page

jesperwiner (changes requested):

almightybruce (approved):
Looks good. I think you should be consistent between `sortOrder` and `sortingOrder`. And maybe `tableHasSavedPersistedSortOrder` can lose the _saved_

jesperwiner (approved):

PR #842, opened 2021-02-25
Feat/drawer avoid close on element data

robinandersson:
Prevents Drawer's close handler on click outside from triggering if element (or parent element) has a specific data-attribute (one default + custom).

### DIFF
--- a/src/Molecules/Drawer/Drawer.tsx
+++ b/src/Molecules/Drawer/Drawer.tsx
@@ -158,15 +158,13 @@ export const Drawer = (React.forwardRef<HTMLDivElement, Props>(
       handleClose();
     }, [handleClose]);
 
-    // TODO: cleanup, fix typescript
     useOnClickOutside(drawerRef, (e) => {
-      const drawerDontCloseOnClickOutside = e.path.some(
-        (target) =>
-          target.dataset?.hasOwnProperty('drawerDontCloseOnClickOutside') &&
-          target.dataset?.drawerDontCloseOnClickOutside,
+      const drawerPreventOnClickOutside = e.nativeEvent.composedPath().some(
+        // to avoid click event target being window or document (= no dataset), check if target is HTMLElement
+        (target) => target instanceof HTMLElement && target.dataset?.drawerPreventOnClickOutside,
       );
 
-      if (closeOnClickOutside && !drawerDontCloseOnClickOutside) {
+      if (closeOnClickOutside && !drawerPreventOnClickOutside) {
         handleClose();
       }
     });
--- a/src/Molecules/Drawer/Drawer.tsx
+++ b/src/Molecules/Drawer/Drawer.tsx
@@ -158,8 +158,15 @@ export const Drawer = (React.forwardRef<HTMLDivElement, Props>(
       handleClose();
     }, [handleClose]);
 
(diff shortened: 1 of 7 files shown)


ratkosolar (approved):

src/Molecules/Drawer/Drawer.tsx
```diff
@@ -158,8 +160,27 @@ export const Drawer = (React.forwardRef<HTMLDivElement, Props>(
       handleClose();
     }, [handleClose]);
 
-    useOnClickOutside(drawerRef, () => {
-      if (closeOnClickOutside) {
+    useOnClickOutside(drawerRef, (e) => {
```
ratkosolar, line 163: not sure if you are aware, but `useOnClickOutside` can take an array of refs as a first argument which should result in similar functionality like the one you built with data attributes but without all the code on the bottom.

although I like how easy the data attributes are to use in the story file :)

toblu (approved):

gustav-pettersson-nordnet (approved):
Like the useOnClickOutside idea as well :)

almightybruce (approved):

src/Molecules/Drawer/Drawer.tsx
robinandersson: Hmm using refs has been considered since it would be too cumbersome – although wasn't aware of the useClickOutside-functionality tbh xD

Either way, the reason we don't want to use refs is since it might be difficult to pass the ref left and right through the app to get it to the drawer (?).

robinandersson: For example, if you want to prevent an element from deep down in the app from closing a specific drawer – this means you would have to pass the ref up all the way up and then down to the drawer. I guess this would require a (new?) context to not cause too overhead, as well as some unavoidable overhead wherever you want to use it.

So yeah, as you say with the data-attribute the code gets a little messy below, but I think it's worth it to avoid overhead for the consumer.

robinandersson: Although much of the logic below is just to make it even easier to use,

E.g. you may safely pass `data-custom-attribute` to both the element and into the drawer's prop – i.e. you don't have to camelcase and remove "-data" yourself (otherwise you would have to pass "customAttribute" in an array to the prop)

robinandersson:
> Like the useOnClickOutside idea as well :)

Gustav are you refering to Ratko's reply? Please take a gander on my replies :)

src/Molecules/Drawer/Drawer.tsx
gustav-pettersson-nordnet: 👍

jesperwiner (approved):

src/Molecules/Drawer/Drawer.tsx
ratkosolar: @robinandersson good points 👍  I also find your solution easier to use, just wanted to shed some light on the useOnClickOutsite as it wasn't very clear it can do that (I learned it myself just recently 🙂 )

robinandersson:
Will merge as soon as this can be QA-tested quickly and thoroughly.

Don't want to risk breaking every drawer in the app 😄

PR #845, opened 2021-03-01
Add support for option groups to single input select

frozzare:


src/Molecules/Input/Select/lib/defaults/ListItem.tsx
```diff
@@ -1,21 +1,27 @@
...
+  const selected = option.options
+    ? false
+    : state.context.selectedItems.includes(option) ||
+      state.context.selectedItems.some(
+        (x) => x.value === option.value || x.options?.some((y: any) => y.value === option.value),
+      );
 
-  return (
+  return option.options ? (
+    <p>{option.label}</p>
```
KitsuneKelso, line 23: Should this be wrapped in `<Typography>` instead? It looks a bit weird with the default font in the input now:
Screenshot 2021-03-01 at 13 39 02

Comparing to the old one, the group labels were bold and italic as well. It feels like the presentation of the groups can be improved.
Screenshot 2021-03-01 at 13 39 23

KitsuneKelso (changes requested):

robineng (approved):

src/Molecules/Input/Select/lib/defaults/ListItem.tsx
frozzare: Yeah, someone forgot to push 🙈

KitsuneKelso (approved):

PR #846, opened 2021-03-01
feat: 🎸 add neutral status to timeline and string type on date

edenvin:
Adding a new status that shows a plain gray circle. This is used for
upcoming events in the timeline. Also extending the date prop to accept
strings. This is also used for upcoming, or ongoing, events where we
don't have a date to show.

### DIFF
--- a/src/Molecules/Timeline/Timeline.tsx
+++ b/src/Molecules/Timeline/Timeline.tsx
@@ -5,6 +5,8 @@ import { Box, Icon, Flexbox, Button, Typography, ListItem, DateTime } from '../.
 
 const getStatusIcon = (status?: string) => {
   switch (status) {
+    case 'NEUTRAL':
+      return <Icon.SolidCircle size={5} fill={(t) => t.color.emptyState} />;
     case 'PENDING':
       return <Icon.InfoPending size={5} fill={(t) => t.color.emptyState} />;
     case 'ACTIVE':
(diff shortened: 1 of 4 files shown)


src/Molecules/Timeline/Timeline.tsx
```diff
@@ -5,6 +5,12 @@ import { Box, Icon, Flexbox, Button, Typography, ListItem, DateTime } from '../.
 
 const getStatusIcon = (status?: string) => {
   switch (status) {
+    case 'NEUTRAL':
+      return (
+        <Icon.IconBase viewBox="0 0 20 20" size={5} fill={(t) => t.color.emptyState}>
+          <path d="M10 20C4.4771525 20 0 15.5228475 0 10S4.4771525 0 10 0s10 4.4771525 10 10-4.4771525 10-10 10z" />
+        </Icon.IconBase>
```
ratkosolar, line 12: I feel like this `Icon.IconBase` goes against the point of compound components for the Icon component, is there any reason for not creating a specific icon for this NEUTRAL state in the `Icon/components` and then exporting that from the `Icon/index.ts` instead?

edenvin: You are right, it felt a bit wonky. Also, after creating a new icon with just a circle I noticed there already is one, that I missed the first time 😅 Using that instead (`SolidCircle`)

KitsuneKelso (approved):

ratkosolar (approved):

PR #856, opened 2021-03-05
Feature/flex table whole row expandable

KitsuneKelso:
The new functionallity is opt-in, supply the new prop `clickRowToExpand` to get the new expandable behaviour.

### DIFF
--- a/src/Molecules/FlexTable/Row/Row.tsx
+++ b/src/Molecules/FlexTable/Row/Row.tsx
@@ -49,6 +49,7 @@ type StyledRowProps = {
   $expanded: boolean;
   $separatorColor: ColorFn;
   $hoverHighlight: boolean;
+  $clickRowToExpand: boolean;
   $xs: ScreenSizeConfigurableProps;
   $sm: Partial<ScreenSizeConfigurableProps>;
   $md: Partial<ScreenSizeConfigurableProps>;
@@ -68,6 +69,8 @@ const StyledRow = styled(Flexbox)<StyledRowProps>`
 
   margin-right: 0;
   margin-left: 0;
+
+  cursor: ${({ $clickRowToExpand }) => ($clickRowToExpand ? 'pointer' : 'auto')};
 
   ${getStylesForSizes<
     {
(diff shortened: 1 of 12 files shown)


nn-davsko (approved):

frozzare (approved):

toblu (approved):
Looks great! 🥇

src/Molecules/FlexTable/Row/Row.tsx
```diff
@@ -107,6 +112,13 @@ const Row: RowComponent = ({
   children,
   ...htmlProps
 }) => {
+  if (clickRowToExpand && 'onClick' in htmlProps) {
```
gustav-pettersson-nordnet, line 115: Nice!

src/Molecules/FlexTable/stories/WithExpandableRows.stories.tsx
```diff
@@ -309,3 +310,105 @@ export const HiddenExpandItemOnDesktop = () => {
...
+          <FlexTable.Cell columnId="column2">Except for on</FlexTable.Cell>
+          <FlexTable.Cell columnId="column3">the chevron...</FlexTable.Cell>
+        </FlexTable.Row>
+      </FlexTable>
+    </div>
+  );
+};
+
+export const ControlledWholeRowExpandalbeTable = () => {
+  // @ts-ignore
```
gustav-pettersson-nordnet, line 373: Why do you need @ts-ignore here?

gustav-pettersson-nordnet (approved):

src/Molecules/FlexTable/stories/WithExpandableRows.stories.tsx
nn-davsko: It's because we couldn't figure out how to explain to ControlledRow that ...rowProps actually contains expand children or expand items which is what it complains about. It is a bit lazy but we to be honest it is a "it's just a story" kind of thing. I will have a look and see if I can make it work in 5 minutes else I might be inclined to just leave it as is if that is agreeable? :)

nn-davsko: More in depth as I can't figure out how to make it happy:

```
export type RowComponents = {
  ExpandItem: ExpandItemComponent;
  ExpandItems: ExpandItemsComponent;
  ExpandRow: ExpandRowComponent;
};
```
ExpandItem, I'm not really sure what it's for? Can we pass in a single item now?

PR #858, opened 2021-03-09
Feat: link underlined

mjmonline:


jesperwiner (approved):

ratkosolar (approved):

src/Molecules/Link/Link.tsx
```diff
@@ -96,12 +101,15 @@ export const Link: LinkComponent = React.forwardRef<any, LinkProps>((props, ref)
...
+        type="button"
         ref={ref}
         className={className}
         onClick={trackClick}
         disabled={disabled}
         as={as}
-        type="button"
+        $color={color}
+        $display={display}
+        $underlined={underlined}
```
ratkosolar, line 112: just curious, what is the convention with `$` prefix?

nn-davsko: the $ denotes a prop that does not propagate past the styled component so it is used for styling purposes but never reaches the <a> tag (where it wouldn't make any sense). I think this is the preferred way to it as opposed to "shouldForwardProp" or whatever it is called.

nn-davsko (approved):

PR #870, opened 2021-03-29
feat: 🎸 add chevron style to accordion item

edenvin:
Adds a prop `withChevron` that uses a chevron to the right instead of a plus/minus to the left as the expand icon in an accordion item. Also allows the title to be a React node instead of only a string.

### DIFF
--- a/src/Molecules/Accordion/__snapshots__/Accordion.stories.storyshot
+++ b/src/Molecules/Accordion/__snapshots__/Accordion.stories.storyshot
@@ -49,8 +49,8 @@ Array [
   width: 100%;
   border: 0;
   padding-top: 12px;
-  padding-right: 0;
   padding-bottom: 12px;
+  padding-right: 0px;
   padding-left: 24px;
   background-color: transparent;
   cursor: pointer;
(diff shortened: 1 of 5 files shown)


fredriklu (approved):

ratkosolar (approved):
maybe I am overthinking it but it would be nicer to have a prop called `style` or `type` or something similar where you can say its "plus-minus" or "chevron" or whatever... I am a bit worried about what would happen when someone decides to add another accordion icon type, then this `withChevron` would need to be refactored... but then again maybe thats a problem we can leave to a person that decides to add that 3rd type 😆  you have my approval 👍

edenvin:
> maybe I am overthinking it but it would be nicer to have a prop called `style` or `type` or something similar where you can say its "plus-minus" or "chevron" or whatever... I am a bit worried about what would happen when someone decides to add another accordion icon type, then this `withChevron` would need to be refactored... but then again maybe thats a problem we can leave to a person that decides to add that 3rd type 😆 you have my approval 👍

Yes, that's a good point! For only two options (as will be the case now) I like this way. I can't foresee any new styles right now, but it can definitely come. I'll leave as it is and we can change it later if that's the case :)

nn-davsko (approved):

PR #884, opened 2021-04-26
feat: 🎸 add scope prop to FeedbackBanner

KitsuneKelso:
dictates the background color of the banner, used depending of scope
placement


anneback (approved):
High quality code

PR #888, opened 2021-04-26
fix: 🐛 styling on ssr when the as prop is used

gustav-pettersson-nordnet:


src/Atoms/Media/Media.tsx
```diff
@@ -42,10 +37,15 @@ const useMedia = (query: string | ((t: Theme) => string)) => {
 
 const Media: React.FunctionComponent<Props> = (props) => {
   const As = props.as || 'div';
+  const StyledAs = styled(As)<{ query: Props['query'] }>`
```
gustav-pettersson-nordnet, line 40: Did not find a way to declare this styled component outside of this component.
Let me know if you know a way of keeping it where it were declared before :)

almightybruce (approved):

src/Atoms/Media/Media.tsx
almightybruce: could you do const StyleAs = (As) => styled(As)><{ query: Props['query'] }> and put that where the StyledDiv was before, and then here do StyledAs = StyleAs(As)?

mhogler (approved):

PR #891, opened 2021-04-28
feat: 🎸 Adds multiple new Icons

NavidNayeri:


src/Atoms/Icon/Icon.stories.tsx
```diff
@@ -19,60 +19,69 @@ export const inlineStory = () => (
   </>
 );
```
anneback, line 21: I removed multiple stories since it uses the knob instead.
Now we have `thinArrow`, `thinChevron` `chevron` and `sortArrow`. And also the stories for the added Icons

anneback (approved):

hardikkaji (approved):

KitsuneKelso (approved):

PR #896, opened 2021-05-04
fix: searchcomponent

victoriatomzik:


jesperwiner (approved):

hardikkaji (approved):

src/Molecules/Input/Select/Select.tsx
```diff
@@ -305,7 +309,9 @@ const Select = (props: Props) => {
               onMouseMove={handleMouseMove}
               ref={listRef}
               data-testid="input-select-list"
-              searchComponent={<SearchWrapper ref={searchRef} component={Search} />}
+              searchComponent={
+                searchComponent || <SearchWrapper ref={searchRef} component={Search} />
```
ratkosolar, line 313: how does the custom search component tie with the functionality? there is no searchRef passed to it

LifeofMichal:
I went through the preview, and I sorta don't get this from UX perspective. 
Is the select component a wrapper for random stuff*** now? Also, how is this different than "On Search Query Change" variant?

*** random stuff being a search bar, I assume ( ??? )

src/Molecules/Input/Select/Select.tsx
LifeofMichal: I'm confused as to the purpose for the searchComponent too, ref aside. We can put in anything in there, what's the useCase?

jesperwiner: Maybe we better just add a prop to not render the SearchWrapper, what we trying to avoid is having a rogue cursor on mobile and popping the virtual keyboard on mobiles

LifeofMichal: Do we need a prop? If the risk is solely on mobile, can we just not load the wrapper on mobile screen size?

ratkosolar: @jesperwiner if that is the case I would suggest you look into disabling autofocus on mobile screens instead of doing this, the logic for that seems to be in this hook -> https://github.com/nordnet/ui/blob/master/src/Molecules/Input/Select/lib/hooks.ts#L25

jesperwiner: But we dont want the keyboard on focus either

src/Molecules/Input/Select/Select.types.ts
```diff
@@ -91,6 +91,7 @@ export type Props = {
    * Use with showSearch only
    */
   onSearchQueryChange?: (e: { type: string; payload: string }) => void;
+  disableSearchComponent?: boolean;
```
ratkosolar, line 94: I am just curious now, how is this different than `showSearch`?

ratkosolar: ok I assume that `showSearch` just makes it visually hidden which might not prevent autofocus from triggering the keyboard on mobile (https://github.com/nordnet/ui/blob/master/src/Molecules/Input/Select/lib/defaults/Search.tsx)... but still adding a new prop for this seems like more of a bandaging then fixing the actual bug, why do we want search field to be visually hidden at all, maybe we should just not render it when its hidden

Issue #899, opened 2021-05-05
Expose some types


LifeofMichal:
Certain pages are importing types from Nordnet/UI by actual folder bath. It is an anti-pattern we possibly should not follow, so we're currently in process of exposing all the types used by webapp-next.

![Screenshot 2021-05-05 at 16 33 23](https://user-images.githubusercontent.com/15144524/117158427-b3183900-adbf-11eb-93fa-90b03f4f3740.png)

There are two relevant branches to the issue, one here, and the other on webapp-next:
- Nordnet/UI: [fix/expose-some-types](https://github.com/nordnet/ui/tree/fix/expose-some-types)
- Webapp-next: [fix/exposing-types-from-nordnet-ui](https://scm.prod.nordnet.se/projects/SHELL/repos/webapp-next/compare/diff?sourceBranch=refs%2Fheads%2Ffix%2Fexposing-types-from-nordnet-ui&targetRepoId=1044)

The branches are being updated in parallel, expose type in UI, update imports in webapp-next, if you're curious to check it out, yalc publish nordnet/ui, add it to webapp-next, and go crazy!

Allegedly that has been on a backburner for some type, so might as well do it now.

victoriatomzik:
@LifeofMichal ready to close?

LifeofMichal:
@victoriatomzik Ah, completely forgot, yes, can be closed :D

PR #900, opened 2021-05-06
Fix/expose FlexTable molecule types

LifeofMichal:
Currently exposed all types from FlexTable molecule, so we can simply import them with simple '@nordnet/ui' instead of specifying the path.

This is not the only that that should be exposed, but it is a good start. Check #899 for more details.

NOTE: since all the types exposed in this issue are being imported directly from their files in webapp-next, deploying this solution is backwards compatible, both '@nordnet/ui' and *path import* will work.


src/index.ts
```diff
@@ -143,6 +164,7 @@ export {
   Modal,
   MultiStepProgress,
   Number,
+  // OnSort,
```
jesperwiner, line 167: Dont we need this ?  we are used to overriding the internal sort with this

jesperwiner (changes requested):

src/index.ts
LifeofMichal: Line 115 same file. Missed the comment before push, my bad.

jesperwiner (approved):

victoriatomzik (approved):

PR #904, opened 2021-05-10
Feat/accordionitem optional hover and padding

frlutz:


src/Molecules/AccordionItem/AccordionItem.tsx
```diff
@@ -6,19 +6,27 @@ import { Box, Icon, Typography } from '../..';
 import { isBoolean, isFunction, isString } from '../../common/utils';
 import { AccordionItemComponent } from './AccordionItem.types';
 
-const Item = styled.div<{ $hasFocus: boolean }>`
+const Item = styled.div<{
+  $hasFocus: boolean;
+  $disableBackgroundColor?: boolean;
```
dskog, line 11: shouldn't this rather be a prop named $backgroundColor which is a colorFn?

frlutz: It could be, but the background color only appears on `hover` or `focus`, passing `backgroundColor` is also a bit misleading. Perhaps `disableHoverAndFocusBackground` is the most correct, but it's a bit to wordy IMO

src/Molecules/AccordionItem/AccordionItem.tsx
```diff
@@ -6,19 +6,27 @@ import { Box, Icon, Typography } from '../..';
...
-    background-color: ${({ theme }) => theme.color.background};
+    background-color: ${({ $disableBackgroundColor, theme }) =>
+      $disableBackgroundColor ? '' : theme.color.background};
   }
 
   outline: ${({ $hasFocus, theme }) => ($hasFocus ? `1px solid ${theme.color.cta}` : 'none')};
 
-  background-color: ${({ $hasFocus, theme }) =>
-    $hasFocus ? `${theme.color.background}` : 'transparent'};
+  ${({ $itemPaddingX, theme }) =>
```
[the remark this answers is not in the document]
frlutz: It's implemented to create distance between everything that `<Item>` has, e.g. the +/-, custom icon or the `chevron`, so that it's not touching the `focus` outline, so therefore I think it's needed here.

Can absolutely change it to `px` though!

nn-davsko (approved):

PR #915, opened 2021-05-28
fix: 🐛 select from filtered options instead of all options

toblu:
When searching options and selecting one of the filtered items, it still selected
from the unfiltered list based on index. This fixes that issue


jesperwiner (approved):

LifeofMichal (approved):
Beautiful!

PR #929, opened 2021-06-14
fix: 🐛 Account for border height in Input phone component

NavidNayeri:


src/Molecules/Input/Select/Select.tsx
```diff
@@ -281,6 +281,7 @@ const Select = (props: Props) => {
           error={error}
           success={success}
           extraInfo={extraInfo}
+          height={props.height}
```
KitsuneKelso, line 284: We should add a story for this new prop.

KitsuneKelso (changes requested):

KitsuneKelso (approved):

anneback (approved):

PR #930, opened 2021-06-15
Bugfix/phone line height chrome

anneback:


NavidNayeri (approved):

src/Molecules/Input/Phone/Phone.tsx
```diff
@@ -214,49 +215,52 @@ const PhoneComponent = React.forwardRef<HTMLInputElement, Props>((props, ref) =>
           onBlur={() => {
             setFocused(false);
           }}
+          disableSearchComponent={disableSearchComponent}
```
anneback, line 218: This is what created the auto focusing on Input.Selects

KitsuneKelso (approved):

src/Molecules/Input/Phone/Phone.tsx
ratkosolar: I remember when they added this property to Input.Select to solve the issue of autofocusing on search field on mobile phones... it makes me wonder, why is that functionality there in a first place?... don't we want to disable this search component always for mobile screens since its not even visible? because otherwise its starting to feel this disableSearchComponent is more of a hack than a solution (or maybe I am missing something here 😄)

PR #944, opened 2021-07-26
Feat/tp 16009 html property passing

denishvachhani:


LifeofMichal (approved):

hardikkaji (approved):

src/Molecules/LinkBuy/LinkBuy.tsx
```diff
@@ -40,6 +41,7 @@ export const LinkBuy: LinkBuyComponent = ({
     to={to}
     target={target}
     rel={rel}
+    {...rest}
```
iamstarkov, line 44: would you mind to be more picky about your arguments?

iamstarkov: passing all of the arguments is usually a bad idea, you need to have a really strong case to actually do it

iamstarkov (changes requested):
please dont spread props

src/Molecules/Drawer/Drawer.stories.tsx
```diff
@@ -362,6 +362,7 @@ export const WithoutCloseOnSpecificElementClick = () => {
           </Box>
         </StyledBoxGreen>
         <Box>
+          {/* eslint-disable-next-line */}
```
victoriatomzik, line 365: why?

LifeofMichal: Lint issue because of no "to" prop, we don't want it to redirect, just to be clicked to illustrate drawer not being closed

src/Molecules/LinkBuy/LinkBuy.tsx
LifeofMichal: We're using this to pass down a custom named html attribute, and we've just noticed that's how its done for both Link and Button, so we followed suit.

LifeofMichal: @iamstarkov The logic is that we can pass a name of an attribute into the Drawer from *wherever it is used, like webapp-next* the name of an attribute that drawer listens to, to not close itself. 

If we start specifying the attribute name now, we have to apply it to the other components, and as well as the Drawer component; it feels like a scope creep for this ticket.

We can go for this, but with other ticket, would that be acceptable for now?

src/Molecules/Drawer/Drawer.stories.tsx
victoriatomzik: then maybe use a Button with an onChange?

LifeofMichal: We're using a Button just below. The point was to show that the Link component can receive the attribute. We technically don't have to have it, since LinkBuy/Sell do the trick to, but we felt it is more illustrative to have it.

iamstarkov (approved):

PR #947, opened 2021-08-02
Fix/tp 16110 add truncation to input select component prop

LifeofMichal:


JohnLindahlTech (approved):
As mentioned in the comment, there is no alternative. Also there is no "actual" change, just an example of usage right?

LifeofMichal:
Ya, exactly, just an example in a story.

We wanted to find a way to implement this pattern into the Select, but weren't sure how to do it without forcing a structure on the CustomSelectedValue.

denishvachhani (approved):

hardikkaji (approved):

PR #953, opened 2021-08-17
feat(input-select): add multi select with grouping

oscarsandberg:
Co-authored-by: Tobias Lundell <lundell.tobias@gmail.com>
Co-authored-by: Oscar Sandberg <oscar.sandberg@nordnet.se>
Co-authored-by: Marcus Hogler <marcus.hogler@nordnet.se>

### DIFF
--- a/src/Molecules/Input/Select/Select.stories.tsx
+++ b/src/Molecules/Input/Select/Select.stories.tsx
@@ -1455,6 +1455,112 @@ export const GroupedOptions = () => {
   );
 };
 
+export const GroupedOptionsMultiselect = () => {
+  const options = [
+    {
+      label: 'Group 1',
+      options: new Array(5).fill(null)?.map((_, i) => ({
+        label: `Child 1 ${i}`,
+        value: `c1-${i}`,
+      })),
+    },
+    {
+      label: 'Group 2',
+      options: new Array(5).fill(null)?.map((_, i) => ({
+        label: `Child 2 ${i}`,
+        value: `c2-${i}`,
+      })),
+    },
+  ];
+
+  const CustomSelectedValue = () => {
+    const [machineState] = useSelectMachineFromContext();
+    const selectedCount = machineState.context.selectedItems.length;
+
+    return (
+      <FlexedBox px={2}>
+        <Typography type="secondary">
+          {selectedCount === 0 ? machineState.context.placeholder : `${selectedCount} selected`}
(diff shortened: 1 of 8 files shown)


src/Molecules/Input/Select/Select.stories.tsx
```diff
@@ -1455,6 +1455,112 @@ export const GroupedOptions = () => {
   );
 };
 
+export const GroupedOptionsMultiselect = () => {
```
nn-davsko, line 1458: Should we consider the two above automated lint warnings as something we should fix before merging this PR? Might help console clutter if nothing else?

src/Molecules/Input/Select/lib/OptionGroup/OptionGroup.tsx
```diff
@@ -0,0 +1,23 @@
...
+  padding-top: ${(p) => p.theme.spacing.unit(p.index === 0 ? 1 : 3)}px;
+  padding-bottom: ${(p) => p.theme.spacing.unit(1)}px;
+  color: ${(p) => p.theme.color.label};
+  pointer-events: none;
+  user-select: none;
+`;
+
+export const OptionGroup: React.FC<{ index: number; label: string }> = ({ index, label }) => {
+  return (
+    <StyledOptionGroup index={index} type="tertiary">
```
nn-davsko, line 19: Is index something Typography understands or should it be transient?

nn-davsko (approved):
Looks good, just a few thoughts might not need any action on your part

src/Molecules/Input/Select/Select.stories.tsx
toblu: Good point, we will try to fix it since we are fiddling with the stories anyway :)

toblu: We looked into this, but don't actually know how to fix it other than adding a fake href prop to `Link`, but that didn't feel like a nice solution because it shouldn't actually link anywhere. So we'll leave it as it is for now

src/Molecules/Input/Select/lib/OptionGroup/OptionGroup.tsx
toblu: Good catch, it's now a transient prop :)

denishvachhani (approved):

PR #956, opened 2021-08-23
Feature/datepicker fullscreen on mobile

nn-davsko:
Makes default and range date pickers render in a full screen modal on mobile sized screens if fullScreenOnMobile prop is passed. Double Date Picker is not part of this feature right now, as far as I can tell it should not be used on small screens.


almightybruce (approved):

robinandersson (approved):

denishvachhani (approved):

toblu (approved):
Nice! 🌟

PR #957, opened 2021-08-24
feat: support of fullScreen in mobile for select element

denishvachhani:
Screenshot 2021-08-24 at 10 29 13

Co-authored-by: David Skog <david.skog@nordnet.se>
Co-authored-by: LifeOfMichal <dzienis.michal.maciej@gmail.com>

### DIFF
--- a/src/Molecules/Input/Select/Select.tsx
+++ b/src/Molecules/Input/Select/Select.tsx
@@ -79,10 +79,9 @@ const Select = (props: Props) => {
     [props.options],
   );
 
-  const smallScreen = useMedia((t) => t.media.lessThan(t.breakpoints.md));
+  const smallScreen = useMedia((t) => t.media.lessThan(t.breakpoints.sm));
 
-  const isFullScreenOnMobileInitialValue =
-    smallScreen && props.fullScreenOnMobileForOptions && !props.withPortal;
+  const isFullscreenOnMobile = smallScreen && props.fullscreenOnMobile && !props.withPortal;
 
   /******      Machine instantiation      ******/
   const machineHandlers = useMachine(SelectMachine, {
@@ -107,8 +106,8 @@ const Select = (props: Props) => {
       uncommittedSelectedItems: [],
       actions: props.actions || [],
       disableSearchComponent: props.disableSearchComponent,
-      fullScreenOnMobileForOptions: isFullScreenOnMobileInitialValue || false,
-      titleOnFullScreen: props.titleOnFullScreen,
+      fullscreenOnMobile: isFullscreenOnMobile || false,
+      titleOnFullscreen: props.titleOnFullscreen,
     },
   });
   const [machineState, send, service] = machineHandlers;
@@ -157,8 +156,8 @@ const Select = (props: Props) => {
       id: props.id,
       actions: props.actions || [],
       disableSearchComponent: props.disableSearchComponent,
-      fullScreenOnMobileForOptions: isFullScreenOnMobileInitialValue || false,
-      titleOnFullScreen: props.titleOnFullScreen,
+      fullscreenOnMobile: isFullscreenOnMobile || false,
+      titleOnFullscreen: props.titleOnFullscreen,
     },
     [
       send,
@@ -176,8 +175,8 @@ const Select = (props: Props) => {
       props.actions,
       props.searchQuery,
       props.disableSearchComponent,
-      isFullScreenOnMobileInitialValue,
-      props.titleOnFullScreen,
+      isFullscreenOnMobile,
+      props.titleOnFullscreen,
     ],
   );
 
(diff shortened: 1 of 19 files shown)


src/Molecules/Input/Select/Select.tsx
```diff
@@ -101,6 +107,8 @@ const Select = (props: Props) => {
       uncommittedSelectedItems: [],
       actions: props.actions || [],
       disableSearchComponent: props.disableSearchComponent,
+      fullScreenOnMobileForOptions: isFullScreenOnMobileInitialValue || false,
```
toblu, line 110: Perhaps the name could be just `fullscreenOnMobile`? That way it would be more aligned with the Datepicker as well as the Modal component

src/Molecules/Input/Select/Select.tsx
```diff
@@ -149,6 +157,8 @@ const Select = (props: Props) => {
       id: props.id,
       actions: props.actions || [],
       disableSearchComponent: props.disableSearchComponent,
+      fullScreenOnMobileForOptions: isFullScreenOnMobileInitialValue || false,
+      titleOnFullScreen: props.titleOnFullScreen,
```
toblu, line 161: I think fullscreen is just one word, so the name should probably be `titleOnFullscreen` (same goes for all other props / components with fullscreen in the name) :)

src/Molecules/Input/Select/Select.tsx
```diff
@@ -337,6 +356,49 @@ const Select = (props: Props) => {
               ))}
             </ListWrapperComponent>
           )}
+          {isOpen && isFullScreenMode && (
+            <Modal open={isOpen} title={titleOnFullScreen}>
+              {/* <FadedScroll> */}
```
toblu, line 361: I guess this was left here on purpose?

src/Molecules/Input/Select/Select.tsx
```diff
@@ -166,6 +176,8 @@ const Select = (props: Props) => {
       props.actions,
       props.searchQuery,
       props.disableSearchComponent,
+      isFullScreenOnMobileInitialValue,
+      props.titleOnFullScreen,
```
toblu, line 180: Have you seen cases in the design where we need to have a separate title for fullscreen? It feels like using just `props.label` would be sufficient for most cases

src/Molecules/Input/Select/lib/wrappers/List.tsx
```diff
@@ -29,12 +30,12 @@ const getStylesForPlacement = (placement: string) => {
 
 const StyledListWrapper = styled.div<any>`
   transform: translate3d(0, 0, 0);
-  position: absolute;
+  position: ${(p) => (p.$fullScreenOnMobileForOptions ? 'none' : 'absolute')};
```
toblu, line 33: I don't think that you can set `position: none;`, none is not a valid position value. If you want to set it to the default value you could use `position: static;`

toblu:
What is the intended behaviour for tablets when having the fullscreen prop? Currently it opens in a modal, but not in fullscreen

Screenshot 2021-08-24 at 09 58 24

src/Molecules/Input/Select/Select.tsx
denishvachhani: Although fullscreenOnMobile is for options, but we renamed it to be in sync with Datepicker

src/Molecules/Input/Select/Select.tsx
denishvachhani: Renamed :)

src/Molecules/Input/Select/Select.tsx
denishvachhani: yeah not needed anymore, removed 👍

src/Molecules/Input/Select/Select.tsx
denishvachhani: checked with design, at this moment, I think prop.lable should be good enough since we don't need custom title for modal.

src/Molecules/Input/Select/lib/wrappers/List.tsx
denishvachhani: good catch, fixed!

toblu (approved):

src/Molecules/Input/Select/lib/MultiSelectList/MultiSelectList.tsx
```diff
@@ -91,34 +100,38 @@ export const Option: React.FC<OptionProps> = ({
...
-        >
-          {label}
-        </EllipsizingText>
-      </FlexboxWidth>
-    </FullHeightFlexbox>
-  </StyledOption>
-);
+  fullscreenOnMobile,
+}) => {
+  return (
```
almightybruce, line 105: minor: no reason to introduce an explicit return

src/Molecules/Input/Select/lib/defaults/ListFullScreen.tsx
```diff
@@ -0,0 +1,76 @@
...
+type ListFullScreenProps = {
+  listPosition?: string;
+  searchComponent?: React.ReactNode;
+  actionsComponent?: React.ReactNode;
+  maxHeight?: string;
+  noFormField?: boolean;
+  placement?: 'bottom' | 'top';
+};
+
+const StyledList = styled(UIList)<any>`
```
almightybruce, line 16: is this meant to be any

almightybruce (approved):

src/Molecules/Input/Select/lib/defaults/ListFullScreen.tsx
denishvachhani: not really :) fixed

mhogler (approved):

src/Molecules/Input/Select/lib/defaults/ListFullScreen.tsx
almightybruce: great! i saw some other anys that you might fix. just do a search :)

denishvachhani: Fixed at one more place, shouldn't be in this file anymore, Though there are several across different files but not newly introduced :)

PR #959, opened 2021-08-26
fix: correct styled components syntax

kk-11:


jesperwiner (approved):

src/Molecules/SfdrBadge/SfdrBadge.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
```
jesperwiner, line 1: this should not be in this PR also maybe make a square prop or shape prop in the existing badge component

src/Molecules/SfdrBadge/SfdrBadge.tsx
```diff
@@ -0,0 +1,48 @@
...
+import styled from 'styled-components';
+import { Typography } from '../..';
+import { Props, colorProps } from './SfdrBadge.types';
+
+const acceptableValues: Array<number | undefined> = [8, 9]; // add article numbers here
+
+const getSfdrColor = (args: colorProps) => {
```
jesperwiner, line 8: this is not what we want in UI, its implementation specific, keep it in webapp-next in a common shared place if possible

jesperwiner (approved):

alvelid (approved):

PR #963, opened 2021-08-30
Fix/input select placeholder color

robinandersson:
Changes `Input` placeholder color to `gray2` (`gray3` in darkmode) according to design to distinguish between selected and placeholder text/options.

### DIFF
--- a/src/theme/theme.types.ts
+++ b/src/theme/theme.types.ts
@@ -230,6 +230,8 @@ export type ThemeColors = {
   paletteMap: string[];
   palettePink: string[];
   paletteTurquoise: string[];
+  /** gray2 */
+  placeholder: RawColor['gray2'];
   /** positive */
   positive: RawColor['positive'];
   /** gray6 */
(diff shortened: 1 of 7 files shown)


mhogler (approved):

denishvachhani (approved):

src/theme/theme.types.ts
```diff
@@ -230,6 +230,8 @@ export type ThemeColors = {
   paletteMap: string[];
   palettePink: string[];
   paletteTurquoise: string[];
+  /** gray2 */
+  placeholder: RawColor['gray2'];
```
almightybruce, line 234: would inputPlaceholder or placeholderInput be more understandable?

almightybruce: i think @jesperwiner had a thought on how the colors should be semantically named?

robinandersson: We choose this name since "placeholder" is such a specific term on frontend, and thus thought "input" was unnecessary (since we have e.g. label as well which isn't specific).

Although we could see the reasoning behind wanting to search for everything input-related. We'll lift this discussion

PR #964, opened 2021-09-03
DatePicker refactoring / improvements / supports fullscreenMode on mobile

denishvachhani:
What has happened

- Big refactoring of Date picker component ( Extracted duplicated logic to `useDatePicker` hook to be used in all type of DatePicker)
- Breaking: Variant is no longer supported, instead use separate component
- `<DatePicker>, <DateRangePicker> and <DoubleDatePicker>`
- DoubleDatePicker is no longer dangerous to use, (Removed warning: use it at your own risk)
-  Keyboard navigation now supported for DoubleDatePicker
-  Add optional Clear button to `<DateRangePicker> and <DoubleDatePicker>`
- FullScreen support in mobile for <DatePicker>, <DateRangePicker> (Check stories for more details)
- Readonly support for Input.Text component
- fixed issue of keyboard trigger on Select component for fullScreenMode
- Breaking: Renamed 'disallowSingleDayRange' prop to 'allowSingleDayRange' prop for `<DateRangePicker> and <DoubleDatePicker>`

Co-Authors: @nn-davsko @toblu 

--------------------------------------------------------------

**DateRangePicker in fullScreenMode for mobile**

Screenshot 2021-09-03 at 11 35 54

--------------------------------------------------------------

**DatePicker in fullScreenMode for mobile**

Screenshot 2021-09-03 at 11 45 04

------------------------------------
**DoubleDatePicker in Desktop**

Screenshot 2021-09-03 at 11 41 49


almightybruce (approved):

src/Molecules/DatePicker/Double/Calendar/DoubleCalendar.tsx
```diff
@@ -156,12 +78,18 @@ const DoubleCalendar: React.FC<Props> = ({
     };
   }, [localeObj, viewedDate]);
 
-  useEffect(() => {
-    if (controlledFocus) setFocused([0, 0]);
-  }, [controlledFocus, setFocused]);
+  const { focusedDate, handleKeyPress, onFocus } = useFocusedDate({
```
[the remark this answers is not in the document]
toblu: Good point, we'll rename it to handleFocus 👍

src/Molecules/DatePicker/shared/components/CalendarDay/CalendarDay.tsx
```diff
@@ -60,67 +78,72 @@ const StyledCalendarDay = styled(Box)<{
...
+    if (selected) return 'buttonText';
+    if (!selected) return 'text';
+    return '';
+  })();
 
-    const handleOnClick = useCallback(() => {
-      if (disabled) return;
-      if (onClick) onClick(date);
-    }, [date, disabled, onClick]);
+  const handleOnClick = useCallback(() => {
```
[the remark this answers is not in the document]
toblu: Agree, will rename!

src/Molecules/DatePicker/shared/hooks/useFocusedDate.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { useOnBlur } from './useOnBlur';
+
+export const useFocusedDate = ({
+  calendarRef,
+  viewedDate,
+  calendar,
+  onMonthChange,
+  dateIsDisabled,
+  onSelectDate,
+}: Props) => {
```
robinandersson, line 15: Consider explicitly typing the return type

src/Molecules/DatePicker/Double/DoubleDatePicker.stories.tsx
```diff
@@ -2,65 +2,50 @@ import { action } from '@storybook/addon-actions';
...
       enableDate={(date) => isSameWeek(dateNow, date)}
       onChange={action('onChange')}
-      variant="DOUBLE"
     />
   );
 };
 
 export const Controlled = () => {
-  const [startDate, setStartDate] = useState(new Date());
+  const [startDate, setStartDate] = useState(new Date('01/01/2021'));
```
robinandersson, line 48: I know it's just a story, so feel free to disregard. But whaddabout setting startOfYear?

src/Molecules/Input/Select/Select.tsx
```diff
@@ -298,7 +295,7 @@ const Select = (props: Props) => {
           id={props.id}
           size={props.size}
           onFocus={handleFocus}
-          onBlur={handleBlur}
+          onBlur={!isFullScreenMode ? handleBlur : undefined}
```
robinandersson, line 298: We usually add the props conditionally right? E.g.

`{ ...( !isFullScreenMode && { handleBlur } ) }`

Although I kinda like this format as well

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
```diff
@@ -71,200 +77,82 @@ const DoubleDatePicker = React.forwardRef<HTMLDivElement, DoubleDatePickerProps>
...
-        }
-        if (selectedStartDate && selectedEndDate && isSameDay(date, selectedEndDate)) {
-          if (disallowSingleDayRange) {
-            return [selectedStartDate, selectedEndDate];
-          }
-          return [selectedEndDate, selectedEndDate];
+        return [selectedDate, selectedDate];
+      }
+      if (selectedEndDate && isSameDay(date, selectedEndDate)) {
+        if (!allowSingleDayRange) {
```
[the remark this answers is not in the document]
toblu: You have a keen eye, sir! 😄  The logic regarding `allowSingleDayRange` is indeed a bit duplicated. We'll do some refactoring to make this better and prevent duplication

src/Molecules/DatePicker/shared/hooks/useDatePicker.types.ts
```diff
@@ -0,0 +1,21 @@
+export type Props = {
+  id: string;
+  selectedDateProp?: Date;
+  selectedEndDateProp?: Date;
+  enableDate?: (date: Date) => boolean;
+  disableDate?: (date: Date) => boolean;
+  onChange?: (date: Date, endDate?: Date | null) => void;
+  onBlur?: (date: Date, endDate?: Date | null) => void;
+  allowDateUpdateOnType?: boolean;
```
robinandersson, line 9: Perhaps an explaining text, why/why not?

src/Molecules/DatePicker/shared/hooks/useDatePicker.tsx
```diff
@@ -0,0 +1,302 @@
...
+  const handleInputSubmit = useCallback(
+    ([startDateValue = inputValue, endDateValue = inputValueEnd] = []) => {
+      const [parsedStartDate, parsedEndDate] = [
+        parseDateString(startDateValue, locale),
+        parseDateString(endDateValue, locale),
+      ];
+
+      const [startDate, endDate] = [allowedDate(parsedStartDate), allowedDate(parsedEndDate)];
+      if (!allowSingleDayRange && startDate && endDate && isSameDay(startDate, endDate)) {
+        return;
```
robinandersson, line 156: Minor: There's some inconsistencies with the usage of if-statements with- and without brackets in conjunction with single line statements

src/Molecules/DatePicker/Range/DateRangePicker.types.ts
```diff
@@ -0,0 +1,59 @@
+export type BaseProps = {
+  ariaLabelPrevious?: string;
+  ariaLabelNext?: string;
+  open?: boolean;
+  allowDateUpdateOnType?: boolean;
```
robinandersson, line 5: Same here, consider adding a description

src/Molecules/DatePicker/Double/DoubleDatePicker.types.ts
```diff
@@ -1,7 +1,8 @@
-export type DoubleDatePickerProps = {
+type DoubleDatePickerProps = {
   ariaLabelPrevious?: string;
   ariaLabelNext?: string;
   open?: boolean;
+  allowDateUpdateOnType?: boolean;
```
robinandersson, line 5: Same here, consider adding description

robinandersson (approved):

src/Molecules/DatePicker/Double/DoubleDatePicker.tsx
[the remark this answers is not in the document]
toblu: Solved now! We also made the `handleSelectDate` logic part of the useDatePicker hook so that we can share it between the range and the double datepicker since they should behave the same way

src/Molecules/DatePicker/shared/hooks/useFocusedDate.tsx
toblu: Hmm, not sure what the benefit of explicitly typing the return would be...? 🤔  It can correctly infer the return type (no `any` types) and letting TS infer the type makes it more flexible to changes since you only need to change in one place. Do you have a case in mind where explicit typing would be beneficial? :)

src/Molecules/Input/Select/Select.tsx
toblu: That's true, we don't usually do it in this format. I also kinda like this format better, it's a bit more explicit what's actually going on. What do you think, should we change it for consistency sake, or should we keep it the way it is for readability sake? I'm leaning towards keeping it this way :)

src/Molecules/DatePicker/Double/DoubleDatePicker.stories.tsx
toblu: The reason why we hard-coded it instead of basing it on today's date, is that the date is included in the storyshots. `startOfYear` would be better than just `new Date()`. But the tests would fail with every new year that we enter. It's not a huge deal of course, but that is the reason why we used a static date string instead :)

src/Molecules/DatePicker/shared/hooks/useDatePicker.tsx
denishvachhani: Nice, removed it since just return without anything

src/Molecules/DatePicker/shared/hooks/useDatePicker.types.ts
denishvachhani: Description added!

src/Molecules/DatePicker/Double/DoubleDatePicker.types.ts
denishvachhani: Description added!

src/Molecules/DatePicker/Range/DateRangePicker.types.ts
denishvachhani: Description added!

src/Molecules/DatePicker/Single/Header/Header.tsx
```diff
@@ -33,48 +34,57 @@ const Header: React.FC<Props> = ({
...
-          </Link>
-        </Box>
-      </Flexbox>
-      <Flexbox container>
+    <Box pb={1}>
+      <Flexbox container justifyContent="space-between">
         <Flexbox item>
-          <SelectMonth id={id} locale={locale} viewedDate={viewedDate} onChange={onMonthChange} />
+          <Box mt={1}>
+            <Link
```
nn-davsko, line 41: yes why aren't these <Button> ?

toblu: Not sure why, they were anchors before. But will try to replace them with buttons instead

nn-davsko (approved):

src/Molecules/DatePicker/Single/Header/Header.tsx
nn-davsko: Agreed it is nothing we changed, I just noticed the github validation message and I don't think there is a case for them being anchors. Worth a shot, it is a bit nitpicky :)

toblu: Yep, buttons seem to work just as well so no reason for anchors like you said. It's good to take the chance and fix these type of things when we are here anyway :) Update incoming!

PR #969, opened 2021-09-10
Disco 5694

alvelid:
Delete "jump scroll to" behaviour when clicking Input.Selects on larger screen sizes. https://jira.prod.nordnet.se/browse/DISCO-5694

### DIFF
--- a/src/Molecules/Input/Select/lib/hooks.ts
+++ b/src/Molecules/Input/Select/lib/hooks.ts
@@ -1,6 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useRef } from 'react';
 import scrollIntoView from 'scroll-into-view-if-needed';
-import { useMedia } from '../../../../..';
+import { useMedia } from '../../../../Atoms/Media';
 import { ContextType } from './context';
 import { Context } from '../machine';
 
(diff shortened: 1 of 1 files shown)


kk-11 (approved):

jesperwiner (approved):

src/Molecules/Input/Select/lib/hooks.ts
```diff
@@ -16,6 +17,7 @@ export const useFocusFromMachine = (
   searchRef: React.RefObject<HTMLInputElement>,
 ) => {
   const isInButtonFocusState = machineState.matches('interaction.enabled.active.focus.button');
+  const isSmallerScreen = useMedia((t) => t.media.lessThan(t.breakpoints.lg));
```
denishvachhani, line 20: Perhaps, you probably want same behaviour for bigger screen table / small desktop. if so you could filter out from smallScreens: i.e  `useMedia((t) => t.media.lessThan(t.breakpoints.md)); `

```javascript
  size: {
    /** Mobile */
    xs: 360;
    /** Tablet, mobile */
    sm: 760;
    /** Tablet, desktop */
    md: 976;
    /** Desktop */
    lg: 1280;
    /** Desktop big */
    xl: 1600;
  };
```

denishvachhani (approved):

src/Molecules/Input/Select/lib/hooks.ts
alvelid: 👍 As I understand we only want this behaviour for bigger screen size, thereby the lg.

PR #976, opened 2021-09-29
Feature/square badge

kk-11:


src/Atoms/Badge/Badge.tsx
```diff
@@ -25,19 +25,19 @@ const animation = css`
...
 `;
 
-const Circle: CircleComponent = styled.span<CircleComponentProps>`
+const Wrapper: WrapperComponent = styled.span<WrapperComponentProps>`
   display: inline-flex;
   justify-content: center;
   align-items: center;
   background-color: ${(p) => (p.backgroundColor ? p.backgroundColor(p.theme) : p.theme.color.cta)};
-  border-radius: ${(p) => p.theme.spacing.unit(p.size)}px;
+  border-radius: ${(p) => (p.variant === 'square' ? 0 : p.theme.spacing.unit(p.size))}px;
```
mjmonline, line 33: maybe better to not have any `border-radius` printed at all, like we do with `color` on line 38

src/Atoms/Badge/Badge.types.ts
```diff
@@ -7,12 +7,14 @@ export type Props = {
   backgroundColor?: ColorFn;
   color?: ColorFn;
   animateOnChange?: boolean;
+  variant?: 'square' | 'circle';
+  weight?: string;
```
mjmonline, line 11: note: this is a prop from Typography component, if that prop type would change then this would be wrong. Maybe better to grab the value of weight directly from the Typography component instead. But then I see other Typography props is being redefined here previously, so 🤷‍♂️

src/Atoms/Badge/__snapshots__/Badge.stories.storyshot
```diff
@@ -1026,6 +1026,89 @@ Array [
...
+      className="c1"
+    >
+      Square Badge
+    </span>
+    <div
+      className="c0"
+    >
+      <span
+        className="c2"
+        size={5}
```
mjmonline, line 1089: Bonus point: You didn't cause this but the component is leaking the `size` prop to this span and that is not a valid span attribute. So this will result in a warning in the console

src/Atoms/Badge/Badge.types.ts
Vikingjonsson: I agree, would be nice to pick it from Typography component, but absolutely no blocker

Vikingjonsson (approved):

PR #1000, opened 2021-11-15
fix: 🐛 add classname to persistant tooltip and related comps

robinandersson:
ClassName wasn't properly passed through for tooltip components, causing
TS error when trying to style specialized components with
styled-components.

### DIFF
--- a/src/Molecules/PersistentTooltip/PersistentTooltip.stories.tsx
+++ b/src/Molecules/PersistentTooltip/PersistentTooltip.stories.tsx
@@ -108,39 +108,6 @@ export const CustomTitleAndDescription = () => {
   );
 };
 
-const StyledPersistentTooltip = styled(PersistentTooltip)`
-  ${PersistentTooltip.components.CloseButtonIcon} {
-    fill: ${(p) => p.theme.color.text};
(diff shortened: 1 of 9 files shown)


src/Molecules/PersistentTooltip/PersistentTooltip.stories.tsx
```diff
@@ -227,6 +194,72 @@ export const PersistentTooltipInDrawer = () => {
   );
 };
 
+const StyledPersistentTooltipWithMargin = styled(PersistentTooltip)`
+  margin-right: 20px;
```
gustav-pettersson-nordnet, line 198: units can be used here

gustav-pettersson-nordnet (approved):

mhogler (approved):

src/Molecules/PersistentTooltip/PersistentTooltip.stories.tsx
robinandersson: Yeah, we discussed this but since this is explicitly done to break away from the design system we opted for using pixels instead

almightybruce (approved):

PR #1006, opened 2021-11-24
Feat/update accordion item

kk-11:


mic1000 (approved):

src/Molecules/AccordionItem/AccordionItem.tsx
```diff
@@ -1,68 +1,67 @@
...
+      ${theme.spacing.unit(pb || py || p || 0)}px ${theme.spacing.unit(pl || py || p || 0)}px;
+
+    & + & {
+      border-top: 1px solid ${theme.color.divider};
+    }
+  `}
 `;
 
-const Button = styled.button<{ $withChevron?: boolean; $disabled?: boolean }>`
+const Button = styled.button<{ withChevron?: boolean; disabled?: boolean }>`
```
frlutz, line 25: Was there any particular reason not to have `withChevron` and `disabled` as transient props?

src/Molecules/AccordionItem/AccordionItem.tsx
```diff
@@ -1,68 +1,67 @@
 import React, { useState } from 'react';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
```
frlutz, line 2: What is the reasoning behind `css` in this file? It seems to only be used in what's already `styled` components, so just returning the complete string instead of `css..` would yield the same behaviour?

src/Molecules/AccordionItem/AccordionItem.tsx
```diff
@@ -1,68 +1,67 @@
...
+      order: ${withChevron ? '1' : '-1'};
+      padding-right: ${withChevron ? 0 : theme.spacing.unit(3)}px;
+    `}
 `;
 
-const OldIconWrapper = styled.div<{ $withChevron?: boolean }>`
-  position: absolute;
-  top: 15px;
-  ${(p) => (p.$withChevron ? 'right' : 'left')}: 0;
+const MovingChevron = styled(Icon.ChevronUp8)<{ disabled?: boolean; expanded: boolean }>`
```
frlutz, line 61: Minor and optional, but "Moving" I don't think it's needed since it's the only type of chevron that should exist right?

frlutz (commented):
Just some minor comments! Otherwise looks good, nice work!

src/Molecules/AccordionItem/AccordionItem.tsx
jesperwiner: think should be put back as $

src/Molecules/AccordionItem/AccordionItem.tsx
jesperwiner: It should work just the same without,  the css function we use when preparing other blocks to be used inside a styled func

jesperwiner (changes requested):
think you could do what Fredrik suggests, its more inline with whats already in UI

src/Molecules/AccordionItem/AccordionItem.tsx
kk-11: In this instance we were previously passing $disabled and disabled as props to the Button (see below). For me generally styled components has been smart enough to remove appropriately named props from being rendered to the dom. For this one though, isn't the desired functionality to have disabled applied to the button? It will still pass as a prop, and now we can remove the extra $disabled being passed without losing the functionality. withChevron is far enough from being anything like an intrinsic attribute that styled components will remove it automatically. Personally I think we should reserve the $ for when it's really needed like expanded on the svg component in this file. It's cool functionality but I would prefer to have p={p} rather than $p={p}. As it kind of confuses what it's supposed to be used for.

jesperwiner: ok

jesperwiner (approved):

PR #1017, opened 2021-12-06
Feat/quick filter

kmelkon:


src/Molecules/QuickFilter/QuickFilter.tsx
```diff
@@ -0,0 +1,115 @@
...
+  & > * {
+    color: inherit;
+  }
+`;
+
+export const QuickFilter: React.FC<Props> = ({
+  label,
+  onChange = () => {},
+  value = '',
+  variant = VARIANT.SMALL,
```
jesperwiner, line 73: I would had called the default one "normal" or "default" or "regular"

src/Molecules/QuickFilter/QuickFilter.stories.tsx
```diff
@@ -0,0 +1,144 @@
...
+      />
+
+      <QuickFilter
+        {...getCardProps()}
+        onChange={onChange}
+        icon={<Icon.Money16 />}
+        label=""
+        variant="big"
+      />
+    </Card>
```
jesperwiner, line 90: please include more variants, like with labels too

src/Molecules/QuickFilter/QuickFilter.tsx
Vikingjonsson: Yeah, that's a good idea. Always hard coming up with a future proof name for these things.
I think we should change it to your suggestion, more generic.

src/Molecules/QuickFilter/QuickFilter.stories.tsx
Vikingjonsson: True, we fix it!

src/Molecules/QuickFilter/QuickFilter.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const FOCUS_RING_SIZES = 2;
+
+const browsersNativeFocusStyles = css`
+  outline: ${FOCUS_RING_SIZES}px solid Highlight;
+  outline: ${FOCUS_RING_SIZES}px solid -webkit-focus-ring-color;
+`;
+
+const StyledInput = styled.input`
+  opacity: 0;
```
kk-11, line 17: What about visibility hidden instead of opacity and pointer events none?

jesperwiner: +1

think we aren't used to these since we been doing so much IE compability, and that one is not IE,  but now we dont have bother with IE any more

Vikingjonsson: It is for accessibility reasons, visibility hidden will remove the screen reader compatibility.
Pointer events: none are because want other events than the default one and to control it  with the div instead of the checkbox

Vikingjonsson (approved):

src/Molecules/QuickFilter/QuickFilter.tsx
jesperwiner: oki

jesperwiner (approved):

alvelid (approved):

PR #1031, opened 2021-12-13
fix: adjust world map colors

mjmonline:


src/theme/theme.types.ts
```diff
@@ -548,6 +548,10 @@ export type ThemeColors = {
   transferStatusBannerText3: string;
   /** index */
   warning: string;
+  /** index */
```
nn-davsko, line 551: this and the one below :)

nn-davsko (approved):

PR #1051, opened 2021-12-21
fix: add husky install to prepare step but don't do it if ci

nn-davsko:


JohnLindahlTech (approved):

package.json
```diff
@@ -29,7 +29,7 @@
     "commit": "git-cz",
     "semantic-release": "semantic-release",
     "prepublishOnly": "yarn build",
-    "postinstall": "if [ -z ${CI} ]; then husky install; else echo 'skip prepare in CI'; fi"
```
nn-davsko, line 32: This is incorrect and strange, there is no postinstall in latest master so the difference would be nothing -> prepare: ...

PR #1057, opened 2022-01-11
Fix firefox dark mode typographies

anneback:


KitsuneKelso (approved):

anneback:
Before:
image

After:
image

src/Molecules/AccordionItem/AccordionItem.tsx
```diff
@@ -137,7 +137,9 @@ export const AccordionItem = React.forwardRef<HTMLButtonElement, Props>(
             disabled={disabled}
             withChevron={withChevron}
           >
-            {title}
```
nn-davsko, line 140: I don't know how we'd like to customise the accordion items, are we sure we always want to have the text as secondary and bold? If not we should probably add a check too see weather the title is a component or just text and in the text case -> wrap it, if component -> just render component. 

But if it is cleared with design then we are good ofc.

nn-davsko (approved):

PR #1063, opened 2022-01-14
Feat/switch toggle component

LifeofMichal:
Just a fancy new toggle based off of the Switch component

Screenshot 2022-01-14 at 15 40 57

### DIFF
--- a/src/Molecules/Switch/SwitchToggle.tsx
+++ b/src/Molecules/Switch/SwitchToggle.tsx
@@ -15,13 +15,13 @@ const Label = styled.label`
   display: inline-block;
 `;
 
-const Knob = styled.span<Pick<SwitchToggleProps, 'knobWidth'>>`
+const Knob = styled.span<Pick<SwitchToggleProps, 'knobwidth'>>`
   background: ${(p) => p.theme.color.bubbleBackground};
   display: block;
   height: ${(p) => p.theme.spacing.unit(KNOB_SIZE)}px;
   width: ${(p) => {
-    if (p.knobWidth) {
-      return p.theme.spacing.unit(p.knobWidth);
+    if (p.knobwidth) {
+      return p.theme.spacing.unit(p.knobwidth);
     }
     return p.theme.spacing.unit(KNOB_WIDTH);
   }}px;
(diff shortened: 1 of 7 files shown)


src/Molecules/Switch/SwitchToggle.tsx
```diff
@@ -0,0 +1,201 @@
...
+const KNOB_WIDTH = 12;
+const TRACK_HEIGHT = TOGGLE_HEIGHT + 1;
+const TRACK_WIDTH = 23;
+
+const Label = styled.label`
+  display: inline-block;
+`;
+
+const Knob = styled.span<Pick<SwitchToggleProps, 'knobWidth'>>`
+  background: white;
```
nn-davsko, line 19: does this not have to be a semantic color if it is to work with themes, or is it always white?

nn-davsko (approved):
looks great, just one question question:
The "bouncy" animation when switching the toggle is nice, but animations are so different in different places, the old switch does not have the bounce but is more linear and boring so I guess I am asking, is this cleared with design?

src/Molecules/Switch/SwitchToggle.tsx
LifeofMichal: oh, my bad, didn't notice; I built on the Switch component, I guess the same issue is there too

LifeofMichal:
> looks great, just one question question: The "bouncy" animation when switching the toggle is nice, but animations are so different in different places, the old switch does not have the bounce but is more linear and boring so I guess I am asking, is this cleared with design?

ya, I showed it off to Lisa, she said it feels "aggressive", but the animation is poached from the same Switch component, so I got no further feedback from her

denishvachhani (approved):

src/Molecules/Switch/SwitchToggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+            role="switch"
+            aria-checked={checked}
+            onClick={isControlled ? onClick : internalClickHandler}
+            disabled={disabled}
+            aria-readonly={readOnly}
+            knobWidth={knobWidth}
+            trackWidth={trackWidth}
+          >
+            <ButtonContent>
+              <Knob knobWidth={knobWidth}>
```
almightybruce, line 157: need transient prop on this?

src/Molecules/Switch/SwitchToggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+                  justifyContent="center"
+                  alignItems="center"
+                  alignContent="center"
+                >
+                  <Typography type="primary" weight="bold" color={(p) => p.color.cta}>
+                    {checked ? valueRight : valueLeft}
+                  </Typography>
+                </Flexbox>
+              </Knob>
+              <Track trackWidth={trackWidth}>
```
almightybruce, line 170: need transient prop?

src/Molecules/Switch/SwitchToggle.tsx
LifeofMichal: Since the Toggle allows putting in custom values apart from currency and %, I added a possibility of adding width to both knob and track to adjust in styled component.

src/Molecules/Switch/SwitchToggle.tsx
LifeofMichal: Follow knob thread

PR #1078, opened 2022-02-02
Feat/icon for all button sizes

stigdanielek:
Button component now includes PillButton as Button.Pill and also the new Button.Icon. Button.Pill is also exported as PillButton for backwards compatibility.


LifeofMichal (approved):

src/Molecules/Button/Button.Gallery.utils.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { Icon } from '../..';
```
nn-davsko, line 2: Possible circular dependency, could you reference the Icon from its own package instead of from an index file? I don't know if Button.Gallery.utils will ever be shared from the same index file but probably best to establish a pattern of going straight towards the components index instead of Molecules

stigdanielek: OK, I will fix this, and some other updates.

jesperwiner (approved):

src/Molecules/Button/components/PillButton/PillButton.stories.tsx
```diff
@@ -2,14 +2,14 @@ import React, { useState } from 'react';
...
 import docs from './PillButton.md';
 
-import { PillButton, Flexbox, Typography, Icon } from '../..';
-import { Display } from '../../common/Display';
+import { Flexbox, Typography, Icon, PillButton } from '../../../..';
+import { Display } from '../../../../common/Display';
 
 export default {
-  title: 'Molecules / PillButton',
+  title: 'Molecules / Button / Button.Pill (PillButton)',
```
almightybruce, line 12: some different size stories would be nice

PR #1096, opened 2022-03-01
fix: 🐛  resolve bug for unwanted scroll to top

abergwall:
Uses scrollTo instead of package to calculate scroll destination.
Prevents unwanted scroll to top when interacting with the page where TabsNav is used.
Co-Authors: William Bruce, Robin Andersson

### DIFF
--- a/src/Molecules/TabsNav/TabsNav.tsx
+++ b/src/Molecules/TabsNav/TabsNav.tsx
@@ -89,7 +89,7 @@ export const TabsNav: Component = ({
     (ref) => {
       if (ref && scrollRef && scrollRef.current && scrollOptions.active) {
         const offsetleft = ref.offsetLeft;
-        const tabWidth = ref.getBoundingClientRect().width;
+        const { width: tabWidth } = ref.getBoundingClientRect();
         const containerWidth = scrollRef.current.getBoundingClientRect().width;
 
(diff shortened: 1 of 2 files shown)


mhogler (approved):

src/Molecules/TabsNav/TabsNav.tsx
```diff
@@ -91,19 +83,26 @@ export const TabsNav: Component = ({
...
   const [setIntersectionLeftRef, intersectionLeftRatio] = useIntersect<HTMLDivElement>();
   const [setIntersectionRightRef, intersectionRightRatio] = useIntersect<HTMLDivElement>();
+  const scrollRef = useRef<HTMLDivElement>(null);
 
   const scrollToRefCallback = useCallback(
     (ref) => {
-      if (ref && scrollOptions.active) {
+      if (ref && scrollRef && scrollRef.current && scrollOptions.active) {
+        const offsetleft = ref.offsetLeft;
+        const tabWidth = ref.getBoundingClientRect().width;
```
almightybruce, line 92: maybe nitpicking but `const { width: tabWidth } = ref.getBoundingClientRect()` would look nicer if you are gonna change anything else :)

denishvachhani (approved):

almightybruce (approved):

PR #1097, opened 2022-03-02
refactor: 💡 export more colors

abergwall:
Exports more colors


src/theme/createDarkColors.tsx
```diff
@@ -37,6 +37,7 @@ export const createDarkColors = (rawColor: RawColors): ThemeColors => {
     badgeBackgroundPositive: rawColor.green500,
     badgeBackgroundWarning: rawColor.yellow400,
     badgeBackgroundNegative: rawColor.red500,
+    badgeBackgroundTurquoise: rawColor.complementaryTurquoise2,
```
jesperwiner, line 40: we try to avoid using names with a color, where is this used ?

abergwall: It will be used for the Badge.Icon in GainAndLossCard

abergwall:
Colors are not needed in new design this PR is closed

PR #1108, opened 2022-03-21
feat: 🎸 option to turn off autocomplete for input number

LifeofMichal:


src/Molecules/Input/Number/Number.types.ts
```diff
@@ -4,6 +4,7 @@ export type Variant = 'normal' | 'quiet';
 
 export type Props = {
   autoFocus?: boolean;
+  autoComplete?: 'on' | 'off';
```
denishvachhani, line 7: since `on` is default value when not set, you probably wouldn't set it explicit ?

LifeofMichal: Ya, I realize it is weird, I had a little bit of a clash of ideas. In input text, right? we have a contract with autoComplete, where you can set it to on, off, email, phone, whatever (its typed as string), so it felt weird to me if I set it as only off here since the prop is so similar to native html "autocorrect". 

I had the prop originally named to autoCorrectOff as bool, but again it would mismatch with the contract in input text.

denishvachhani (approved):

hardikkaji (approved):

src/Molecules/Input/Number/Number.types.ts
almightybruce: Why not use that same contract to allow all html variations?

LifeofMichal: Because this is a specialized field, and one would most likely not have to send in any additional types of autocompletes other than the ones required by numbers? Maybe? Is that a good argument?

jesperwiner (approved):

almightybruce (approved):

PR #1114, opened 2022-03-28
fix: statusmodal - add argument in onclose to be able to determine which button was pressed

gustafsorensen:


johcal (dismissed):

src/Organisms/StatusModal/StatusModal.tsx
```diff
@@ -7,10 +7,13 @@ const StatusModal: React.FC<Props> = ({ id = '', loading = false, onClose, optio
 
   const { status, title, text, textConfirm, textCancel } = options || {};
 
-  const closeModal = useCallback(() => {
-    if (onClose) onClose();
-    setShowModal(false);
-  }, [onClose, setShowModal]);
+  const closeModal = useCallback(
+    (confirmed) => {
```
jesperwiner, line 11: I know it's old stuff but if you wanna you can improve it with the isFunction check and some { } 

import { isFunction } from '../../common/utils';

      if (isFunction(onClose)) {
        onClose(confirmed);
      }

gustafsorensen: Added your suggestion 👍

johcal (approved):

jesperwiner (approved):

PR #1119, opened 2022-04-11
Feature/full width title collapsible card

jonnakollin:
Co-Author: Tobias Thun [tobias.thun@nordnet.se](mailto:tobias.thun@nordnet.se)

### DIFF
--- a/src/Molecules/CollapsibleCard/CollapsibleCard.tsx
+++ b/src/Molecules/CollapsibleCard/CollapsibleCard.tsx
@@ -57,9 +57,9 @@ const AnimatedChevronUp = styled(OldIcon.ChevronUp)<Pick<IndicatorsProps, '$coll
   transition: transform 0.16s ease-out;
 `;
 
-const StyledFlexboxItem = styled(Flexbox)<{ fullWidthTitle: boolean | undefined }>`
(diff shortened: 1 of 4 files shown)


nn-davsko (approved):

src/Molecules/CollapsibleCard/CollapsibleCard.tsx
```diff
@@ -57,6 +57,14 @@ const AnimatedChevronUp = styled(OldIcon.ChevronUp)<Pick<IndicatorsProps, '$coll
   transition: transform 0.16s ease-out;
 `;
 
+const StyledFlexboxItem = styled(Flexbox)<{ fullWidthTitle: boolean | undefined }>`
+  ${(props) =>
```
emilwestin, line 61: I think it would be a bit more readable to extract 'fullWidthTitle' from props, e.g. ${( { fullWidthTitle } ) => ... }. I also don't think you need to type it as 'boolean | undefined'. It should be enough with just boolean?

jonnakollin: It's fixed now. :)

emilwestin (approved):

PR #1123, opened 2022-04-20
Feat/fix progress bar

kk-11:


src/theme/createDarkColors.tsx
```diff
@@ -353,6 +353,12 @@ export const createDarkColors = (rawColor: RawColors): ThemeColors => {
     worldMapLand: rawColor.blue600, // not yet in use
     worldMapWater: rawColor.gray800, // not yet in use
 
+    removeFundText: rawColor.red500,
```
jesperwiner, line 356: I would had maybe generalized it a bit,   removeText.  or deleteText.   so when next one of this comes it doesnt feel wrong to use for a ETF remove button text

src/theme/createDarkColors.tsx
```diff
@@ -353,6 +353,12 @@ export const createDarkColors = (rawColor: RawColors): ThemeColors => {
...
+
+    searchFundPopupBackground: rawColor.blue800,
```
jesperwiner, line 358: same here.  maybe. searchPopupBackground is good enough

jesperwiner (approved):

Vikingjonsson (approved):

src/theme/createDarkColors.tsx
Vikingjonsson: Nice fix for the variable names in the update

PR #1140, opened 2022-05-09
Fix/color correction

mhogler:


mjmonline (approved):

src/Molecules/FeedbackBanner/FeedbackBanner.tsx
```diff
@@ -14,11 +14,11 @@ const getBorderColor = ({
 }) => {
   switch (variant) {
     case 'error':
-      return theme.color.negative;
+      return theme.color.functionRed;
```
LifeofMichal, line 17: are those really the names you people came up with?

src/Molecules/FormField/FormField.tsx
```diff
@@ -141,7 +141,7 @@ export const FormField = React.forwardRef<HTMLDivElement, Props>(
               exit={{ y: 0, opacity: 0 }}
               aria-live="polite"
             >
-              <Typography type="tertiary" color={(t) => t.color.negative}>
+              <Typography type="tertiary" color={(t) => t.color.danger}>
```
LifeofMichal, line 144: imindanger

LifeofMichal: What's the difference between functionRed and Danger colours? Conceptually, I don't mean the hex code

LifeofMichal (approved):

src/Molecules/FeedbackBanner/FeedbackBanner.tsx
mhogler: They are called function colors, just like brand is called brand colors etc

Screenshot 2022-05-09 at 16 39 57

jesperwiner: why not call it theme.color.error ?  if you want to avoid reusing positive and negative

having colors named the color is what we try avoid ? if at all possible, i know there is exceptions (i probably added a few)

src/Molecules/FormField/FormField.tsx
mhogler: FunctionRed is supposed to be the same color in light/dark mode, whereas danger will be slightly different  shade between the two. Danger will mostly be used for error input fields, and functionRed for 'shapes' such as icons etc

src/Molecules/FeedbackBanner/FeedbackBanner.tsx
mhogler: Yeah you're right @jesperwiner, we'd like to avoid having the color name in the color variable. However, the "functionRed" is not for errors, it's for all types of shapes where we now use negative (for instance on icons).
We named it function because that's what design call it, I'm unsure of the purpose of inventing another name and having the ui color names differ from figma

jesperwiner: alright

jesperwiner (approved):

edvber (approved):

src/Molecules/FeedbackBanner/FeedbackBanner.tsx
robinandersson: We use brandPink, brandTurqoise etc. as names for default colors already, so I think this is fine 🙏

PR #1141, opened 2022-05-11
Refactor/rounded pill

kmelkon:


emilwestin:
We had the same issue with all snapshots failing. Solved it by doing a `yarn install` prior to running `yarn test` :)

kk-11 (approved):

PR #1148, opened 2022-05-25
feat: 🎸 Round checkbox

edvber:
Adds round option to input checkbox, TP-18383

### DIFF
--- a/src/Molecules/Input/Checkbox/Checkbox.stories.tsx
+++ b/src/Molecules/Input/Checkbox/Checkbox.stories.tsx
@@ -26,10 +26,12 @@ defaultStory.story = {
   name: 'Default',
 };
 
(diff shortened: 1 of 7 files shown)


src/Molecules/Input/Checkbox/Checkbox.stories.tsx
```diff
@@ -26,6 +26,12 @@ defaultStory.story = {
   name: 'Default',
 };
 
+export const roundStory = () => <Input.Checkbox name="example" value="green" label="Green" round />;
```
emilwestin, line 29: An idea for "scalability" would be to have a prop called `shape` or something similiar that could take the value round, instead of having it as a boolean.

almightybruce (dismissed):

emilwestin (approved):

almightybruce (approved):

PR #1149, opened 2022-05-27
fix: 🐛 don't render hidden search input on mobile screen

emilwestin:
Input.Select no longer renders a hidden text input on small screens. On
mobile devices the virtual keyboard would appear and auto focus on the
input element. Also adds TabIndex when noFormField prop is set due to
inconsistent focus and blur behaviors between safari and other browsers.

### DIFF
--- a/src/Molecules/Input/Select/Select.tsx
+++ b/src/Molecules/Input/Select/Select.tsx
@@ -84,7 +84,7 @@ const Select = (props: Props) => {
   const isFullscreenOnMobile = smallScreen && props.fullscreenOnMobile && !props.withPortal;
 
   const isDisableSearchComponent =
-    isFullscreenOnMobile && !props.showSearch ? true : props.disableSearchComponent;
+    smallScreen && !props.showSearch ? true : props.disableSearchComponent;
 
   /******      Machine instantiation      ******/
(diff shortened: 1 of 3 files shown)


alvelid (approved):

src/Molecules/Input/Select/Select.tsx
```diff
@@ -84,7 +84,7 @@ const Select = (props: Props) => {
   const isFullscreenOnMobile = smallScreen && props.fullscreenOnMobile && !props.withPortal;
 
   const isDisableSearchComponent =
-    isFullscreenOnMobile && !props.showSearch ? true : props.disableSearchComponent;
+    smallScreen && !props.showSearch ? true : props.disableSearchComponent;
```
ratkosolar, line 87: shouldn't it be smallScreen || ?
or maybe I am misinterpreting the intention here :D

gabagoo: If you set showSearch={true} on the component you don't want to disable search

gabagoo: If you set showSearch={true} on the component you don't want to disable search

ratkosolar (approved):

PR #1157, opened 2022-06-20
Feature/add delay to skeleton

abergwall:


src/Atoms/Skeleton/Skeleton.tsx
```diff
@@ -21,6 +21,14 @@ const getSize = (x: StringOrNumber, theme: Theme) => {
   return isNumber(x) ? `${theme.spacing.unit(x)}px` : x;
 };
 
+const getDelay = (delay?: boolean | number) => {
```
robinandersson, line 24: minor:

"get" keyword feels like an external data fetching IMO, maybe "determine" or something similar would be preferable+

src/Atoms/Skeleton/Skeleton.tsx
```diff
@@ -21,6 +21,14 @@ const getSize = (x: StringOrNumber, theme: Theme) => {
...
+  if (typeof delay === 'number') {
+    return delay;
+  }
+
+  return delay === false ? 0 : 1000;
```
robinandersson, line 29: This makes delay set to 1000 by default (if prop is omitted) right, was this intended? (personally I think the delay should be on by default)

If intended it should probably be documented in e.g. the types-declaration 🙏

robinandersson (dismissed):

src/Atoms/Skeleton/Skeleton.tsx
robinandersson: Should the length be 1 second by default though? I think there's a specific design guideline for what's an appropriate delay for best UX

abergwall: Copying the behaviour of the loading spinner so it will be false by default. Will add docs to types declaration

edvber (approved):

src/Atoms/Skeleton/Skeleton.tsx
```diff
@@ -52,20 +60,34 @@ const Circle = styled(Base)`
...
   };
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      if (noDelay) {
+        return;
+      }
+      setRenderSkeleton(true);
+    }, deriveDelayFromProps(delay));
+    return () => clearTimeout(timer);
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
```
gabagoo, line 79: Why not just have exhaustive deps here?

gabagoo (approved):

src/Atoms/Skeleton/Skeleton.tsx
LifeofMichal: I guess this is supposed to be run only once at startup?

src/Atoms/Skeleton/Skeleton.tsx
```diff
@@ -52,20 +60,34 @@ const Circle = styled(Base)`
...
 
-  if (variant === 'rect') {
+  if (variant === 'rect' && (noDelay || renderSkeleton)) {
```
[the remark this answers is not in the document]
abergwall: true

almightybruce (approved):

PR #1174, opened 2022-08-30
feat: 🎸 add util function to get units

emilwestin:
when consuming the ui librabry, this util abstracts away the need to
call the theme when you need units. Instead of writing ${(theme) =>
theme.spacing.unit(x)}px, you can use the util and write ${units(x)}px;
instead

### DIFF
--- a/src/common/unitUtils.ts
+++ b/src/common/unitUtils.ts
@@ -1,0 +1,7 @@
+import { Theme } from '../theme/theme.types';
+
+export const units =
+  (unit: number) =>
+  ({ theme }: { theme: Theme }) => {
+    return theme.spacing.unit(unit);
+  };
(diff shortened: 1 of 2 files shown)


hardikkaji:
@emilwestin just wondering how import statement looks like for this?

emilwestin:
@hardikkaji, It would look like the import for everything else in the ui kit:

`import { units } from '@nordnet/ui'; `

hardikkaji:
> @hardikkaji, It would look like the import for everything else in the ui kit:
> 
> `import { units } from '@nordnet/ui'; `

Looks good! I thought it would be parts of some utils! So just got curious! :)

hardikkaji (approved):

mhogler (approved):

denishvachhani (approved):

PR #1181, opened 2022-09-16
Feat/new graph area icon and mobile styled toggle

LifeofMichal:


jesperwiner (changes requested):
I think there must be a better way of updating the existing Switch component instead of a copy and small tweaks to it.

src/Molecules/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,137 @@
...
+const TRACK_HEIGHT = 5;
+const TRACK_WIDTH = 9;
+const KNOB_SIZE = 4;
+
+const Label = styled.label`
+  display: inline-block;
+`;
+
+const Knob = styled.span`
+  background: ${(p) => p.theme.color.bubbleBackground};
```
jesperwiner, line 17: this is also wrong in the Switch component but no one gotten around to yet, messes up dark mode disabled state

jesperwiner: I've asked design since the colors in Figma are not correct for disabled variant in dark mode

PR #1203, opened 2022-10-20
fix: add pointer cursor to control-list-item

jesperwiner:


src/Molecules/ControlsListItem/ControlsListItemSelect.tsx
```diff
@@ -63,7 +63,7 @@ const ControlsListItemSelect: React.FC<Props> = ({
   }
 
   return (
-    <ControlsListItem>
+    <ControlsListItem disabled={disabled}>
```
kk-11, line 66: I think disabled has the potential to bleed to the dom. Might be good to use transient prop

kk-11 (approved):

kmelkon (approved):

PR #1234, opened 2022-11-14
feat: React 18

mjmonline:


iamstarkov (approved):
🫣

src/Atoms/Box/Box.types.ts
```diff
@@ -55,7 +55,7 @@ export type Props = {
    * Render this element as different DOM node
    * @default div
    * */
-  as?: keyof JSX.IntrinsicElements | React.ComponentType<any>;
+  as?: any;
```
denishvachhani, line 58: intended?

LifeofMichal: shouldn't it be something like HTML.Element<any> at least?

src/Atoms/Flexbox/Flexbox.types.ts
```diff
@@ -1,13 +1,15 @@
...
 export type Width = number | string;
 export type Height = number | string;
 
 export type ContainerProps = {
+  children?: React.ReactNode;
   container?: boolean;
   width?: Width;
   height?: Height;
-  as?: keyof JSX.IntrinsicElements | React.ComponentType<any>;
+  as?: any;
```
LifeofMichal, line 12: same comment as in Box.types.ts

src/Atoms/Box/Box.types.ts
mjmonline: yes it's intended, we had a lots of issues and we didn't want to get stuck at this TS error. It's being compared to `string | ComponentType<any> | undefined` which is a bit problematic. On the positive side all `as` props now has the same type so it would be easy to search and replace

mjmonline: @LifeofMichal I get this with the suggested type "Type 'HTMLElement' is not generic."

src/Atoms/Flexbox/Flexbox.types.ts
mjmonline: all `as` props is set to `any` with this PR

iamstarkov (dismissed):

iamstarkov (dismissed):

PR #1238, opened 2022-11-15
fix: need to declare ref on tooltip

nn-davsko:
Motivation is this: 

`TS2322:` Type '{ children: Element; ref: MutableRefObject<undefined>; isOpen: boolean; label: Element; pointerEvents: true; }' is not assignable to type 'IntrinsicAttributes & Props'.
  Property 'ref' does not exist on type 'IntrinsicAttributes & Props'.`


nn-davsko:
I need to admit though, I have not found documentation regarding refs obeying the same rules as children in React 18 i.e. no more implicit part of React.FC

PR #1335, opened 2023-04-03
chore: 🤖 allow node 19.x as engine

adanyb:


adanyb:
Wait for Node 19 LTS before merging this

PR #1384, opened 2023-07-13
fix: 🐛 use set functions for border-radius

jonnakollin:
use set functions for border-radius to prevent snowflaking


src/theme/theme.ts
```diff
@@ -81,7 +81,11 @@ export const createTheme = (config: ThemeConfig = {}): Theme => {
     gutter: GUTTER,
   };
 
-  const borderRadius = (radius: BORDER_RADIUS) => `${radius}px`;
+  const borderRadius2 = () => '2px';
```
LifeofMichal, line 84: I know I suggested those could be functions, but it feels like those could just be consts, so we can just have theme.borderRadius2 without functions :P

LifeofMichal (approved):

alvelid (approved):

PR #1398, opened 2023-08-14
feat: upgrade to styled-components v6

mjmonline:


src/Molecules/DatePicker/shared/components/CalendarDay/CalendarDay.tsx
```diff
@@ -131,6 +131,8 @@ const CalendarDay: React.FC<CalendarDayProps> = ({
       $isFirstDay={isFirstDay}
       $isLastDay={isLastDay}
       data-testid={`${format(date, 'MMMM d', { locale })}`}
+      data-selected={selected}
+      data-disabled={disabled}
```
mjmonline, line 135: I added these because tests were failing because styled-components now uses stylis. Normaly we would have a button component or something that natively is disabled or has aria-selected but because this is just a div I thought it was better to test against these attribute rather than to try to fix tests testing styles

jesperwiner (approved):

gustav-pettersson-nordnet (approved):

PR #1402, opened 2023-08-16
fix: patch breaking change

gustafsorensen:


mjmonline (changes requested):
let's try to solve the problem instead. You will have to do these if's in a lot of places

PR #1415, opened 2023-09-07
fix: no arrow when rounded corners is enabled

gustav-pettersson-nordnet:


src/Atoms/DropdownBubble/DropdownBubble.tsx
```diff
@@ -135,5 +135,7 @@ export const DropdownBubble = styled.div<Props>`
     )};
   border-radius: ${(p) => p.theme.borderRadius4};
   ${bottomAndTopPlacementCss}
-  ${triangleCss}
+
+  ${({ theme, triangle }) =>
+    triangle || theme.isFeatureEnabled('roundedCorners') === false ? triangleCss : ''}
```
emilwestin, line 140: Perhaps this is more readable if you switch the if statement and check if the feature is enabled? 

I guess it's something like this: `!triangle || theme.isFeatureEnabled('roundedCorners') ? '' : triangleCss`
It feels a little weird to check if its not enabled

mjmonline (dismissed):

mjmonline (approved):

PR #1423, opened 2023-09-14
feat: add borderRadius function to theme and adjust barscale

mjmonline:


src/theme/theme.ts
```diff
@@ -92,6 +92,8 @@ export const createTheme = (config: ThemeConfig = {}): Theme => {
   const RADIUS_LARGE: BORDER_RADIUS = 20;
   const RADIUS_100: BORDER_RADIUS = 100;
 
+  const borderRadius: Theme['borderRadius'] = (x) =>
```
emilwestin, line 95: Should `x` be typed to be a `BORDER_RADIUS` so we "lock" what type of radius you are allowed to input?

emilwestin: Or is the purpose of this the be able to override and do sent in whatever you like?

emilwestin: Nvm.. see that you typed it below..! Disregard this :D

mjmonline: it is already when we type the whole function to `Theme['borderRadius']`

emilwestin (approved):

PR #1441, opened 2023-10-21
feat: 🎸 Adding new avatar component prop

patricksjoberg:
This would makie it possible to send a react component as prop instead of image or icon. Maybe a little breaking change but I will make a PR changing the other components as well.

### DIFF
--- a/src/Atoms/Box/Box.types.ts
+++ b/src/Atoms/Box/Box.types.ts
@@ -26,7 +26,7 @@ type Margins = {
   my?: number | string;
   m?: number | string;
 };
-type Paddings = {
+export type Paddings = {
   pt?: number | string;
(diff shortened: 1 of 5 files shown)


src/Atoms/Box/Box.types.ts
```diff
@@ -26,7 +26,7 @@ type Margins = {
   my?: number | string;
   m?: number | string;
 };
-type Paddings = {
+export type Paddings = {
```
patricksjoberg, line 29: I don't know if this is allowed?

davthoNordnet: It should be fine imo!

davthoNordnet (approved):

PR #1470, opened 2023-11-01
feat: 🎸 show cursor on hover

patricksjoberg:
Toggle component will have a cursor on hover unless it is disabled.

### DIFF
--- a/src/Molecules/Toggle/Toggle.tsx
+++ b/src/Molecules/Toggle/Toggle.tsx
@@ -9,11 +9,12 @@ const TRACK_HEIGHT = { s: 4, m: 5, l: 4 };
 const TRACK_WIDTH = { s: 7, m: 9, l: 10 };
 const KNOB_SIZE = { s: 3, m: 4, l: 6 };
 
-const Label = styled.label`
+const Label = styled.label<{ disabled?: boolean }>`
   display: inline-block;
+  cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
 `;
 
-const Knob = styled.span<{ $size: SizeProp }>`
+const Knob = styled.span<{ $size: SizeProp; disabled?: boolean }>`
   background: ${(p) => p.theme.color.toggleKnobEnabledOffBg};
   display: block;
   height: ${(p) => p.theme.spacing.unit(KNOB_SIZE[p.$size])}px;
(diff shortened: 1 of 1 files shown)


src/Molecules/Toggle/Toggle.tsx
```diff
@@ -9,11 +9,12 @@ const TRACK_HEIGHT = { s: 4, m: 5, l: 4 };
 const TRACK_WIDTH = { s: 7, m: 9, l: 10 };
 const KNOB_SIZE = { s: 3, m: 4, l: 6 };
 
-const Label = styled.label`
+const Label = styled.label<{ disabled?: boolean }>`
```
mjmonline, line 12: you probably want to say `$disabled?: boolean` instead. Otherwise styled-components will pass the prop through to the `<label />` tag and disabled is not a valid attribute on that tag

src/Molecules/Toggle/Toggle.tsx
```diff
@@ -9,11 +9,12 @@ const TRACK_HEIGHT = { s: 4, m: 5, l: 4 };
...
   display: inline-block;
+  cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
 `;
 
-const Knob = styled.span<{ $size: SizeProp }>`
+const Knob = styled.span<{ $size: SizeProp; disabled?: boolean }>`
```
mjmonline, line 17: same

src/Molecules/Toggle/Toggle.tsx
```diff
@@ -26,16 +27,18 @@ const Knob = styled.span<{ $size: SizeProp }>`
   box-sizing: border-box;
   box-shadow: 0px 1px 3px 1px ${(p) => p.theme.color.shadowSwitch};
   transition: transform 0.2s cubic-bezier(0.18, 0.9, 0.35, 1.15);
+  cursor: ${(props) => (props.disabled ? 'not-allowed' : 'pointer')};
 `;
 
-const Track = styled.span<{ $size: SizeProp }>`
+const Track = styled.span<{ $size: SizeProp; disabled?: boolean }>`
```
mjmonline, line 33: same

mjmonline (changes requested):

src/Molecules/Toggle/Toggle.tsx
patricksjoberg: Ah thanks! 👍

mjmonline (approved):

PR #1475, opened 2023-11-06
[TP-23142] Add thousand separator to number input

TahirMehdi:
Add thousand separator to number input
Add getLocaleStringAsNumber and getNumberAsLocaleString utils Fix input caret jumping issue
Add tests

### DIFF
--- a/src/Molecules/Input/Number/Number.stories.tsx
+++ b/src/Molecules/Input/Number/Number.stories.tsx
@@ -78,6 +78,53 @@ withDefaultValueUncontrolledBehaviour.story = {
   name: 'With default value (Uncontrolled behaviour)',
 };
 
+export const withThousandSeparatorValueControlled = () => {
+  const Component = () => {
+    const [value, setValue] = useState(10);
+    const changeHandler = (v: string) => {
+      const newValueAsNumber = parseInt(v, 10);
+      setValue(newValueAsNumber);
+    };
+
+    return (
+      <>
+        <Input.Number
+          id="insert-unique-id"
+          label="Label"
+          value={value}
+          onChange={changeHandler}
+          withThousandSeparator
+        />
+        <button type="button" onClick={() => setValue(value - 1)}>
+          Decrease
+        </button>
+        <button type="button" onClick={() => setValue(value + 1)}>
+          Increase
+        </button>
+      </>
+    );
+  };
+  return <Component />;
+};
+
+withThousandSeparatorValueControlled.story = {
+  name: 'With thousands separator (Controlled behaviour)',
+};
+
+export const withThousandSeparatorValueUncontrolled = () => (
+  <Input.Number
+    id="insert-unique-id"
+    label="Label"
+    defaultValue="15.2"
+    step="0.1"
+    withThousandSeparator
+  />
+);
+
+withThousandSeparatorValueUncontrolled.story = {
+  name: 'With thousands separator (Uncontrolled behaviour)',
+};
+
 export const withASmallerStep = () => (
   <Input.Number id="insert-unique-id" label="Label" defaultValue="15.200" step="0.005" />
 );
--- a/src/Molecules/Input/Number/Number.tsx
+++ b/src/Molecules/Input/Number/Number.tsx
@@ -1,14 +1,20 @@
-import React, { useRef, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { useIntl } from 'react-intl';
 import styled, { css } from 'styled-components';
 import * as R from 'ramda';
-import { NumberComponent, Props, Variant } from './Number.types';
 import { Flexbox, FormField, OldIcon, Typography, VisuallyHidden } from '../../..';
 import NormalizedElements from '../../../common/NormalizedElements';
-import { getNumberAsString, getStringAsNumber } from './utils';
-import { assert, isNumber, isString, isUndefined } from '../../../common/utils';
-import adjustValue from './adjustValue';
 import { placeholderNormalization } from '../Text/Text';
+import { assert, isNumber, isString, isUndefined } from '../../../common/utils';
+import {
+  getLocaleStringAsNumber,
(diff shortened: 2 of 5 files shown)


edvber (dismissed):

src/Molecules/Input/FormattedNumber/FormattedNumber.styles.tsx
```diff
@@ -0,0 +1,190 @@
...
+  ${focusBorderStyles}
+  border-width: ${(p) => (p.variant === 'quiet' ? '0 0 2px 0' : '1px')};
+
+  &:focus {
+    border-width: 1px;
+  }
+  ${(p) =>
+    p.disabled && p.variant === 'quiet'
+      ? `border-color: ${
+          p.theme.isDarkMode ? p.theme.color.inputBorder : p.theme.color.buttonBackgroundDisabled
```
jesperwiner, line 70: any reason not changing to colorTokens ?

TahirMehdi: Styling was copied from Number component, and as FormatNumber took a while to implement, we decided to omit other changes except for functionality ones.

src/Molecules/Input/FormattedNumber/utils.ts
```diff
@@ -0,0 +1,5 @@
+import { Props } from './FormattedNumber.types';
+
+/** True if error is non-empty string */
+export const hasError = (error?: Props['error']): boolean | undefined =>
+  error === undefined ? undefined : error.length > 0;
```
mjmonline, line 5: This function is kinda hard to read. Can we separate the TS to a line above? Something like `type HasError = (error?: Props['error']) => boolean | undefined;` Or maybe not using ternery ifs? One of the two maybe

src/Molecules/Input/FormattedNumber/FormattedNumber.styles.tsx
mjmonline: I have to agree with @jesperwiner, we should avoid adding stuff we need to replace

src/Molecules/Input/FormattedNumber/FormattedNumber.styles.tsx
mjmonline, line 1: I know you didn't write this file but boy does it need some cleaning up. Not to mension the leaking props. We can do this in a future iteration though

src/Molecules/Input/FormattedNumber/FormattedNumber.tsx
```diff
@@ -0,0 +1,205 @@
...
+          'hideLabel',
+          'label',
+          'labelTooltip',
+          'labelTooltipPosition',
+          'labelTooltipInModal',
+          'width',
+          'className',
+        ],
+        props,
+      )}
```
mjmonline, line 122: Ramda is not playing nice with TS, specially in the branch where we update ts or linting packages. Would love if you could do this with something like `{ x, y, z, ...formattedNumberProps } = props`

tsconfig.json
```diff
@@ -25,5 +25,5 @@
     "ignoreDeprecations": "5.0" // TODO: remove suppressImplicitAnyIndexErrors and this line
   },
   "include": ["src"],
-  "exclude": ["node_modules", "build", "scripts", ".history", "**/*.test.ts*"]
```
mjmonline, line 28: Could we get a short description of what drove this change? Not against it, just curious

src/Molecules/Input/FormattedNumber/FormattedNumber.tsx
```diff
@@ -0,0 +1,205 @@
...
+          )}
+        </FormattedNumberComponents.Wrapper>
+      </Typography>
+    </FormField>
+  );
+};
+FormattedNumberInput.components = FormattedNumberComponents;
+export const FormattedNumber: React.FC<Props> & {
+  components: typeof FormattedNumberComponents;
+} = FormattedNumberInput as any;
```
mjmonline, line 205: We are de defining the types twice, once on line 11 to 22 and then here. Also in the end we cast it all to `any` hehe

package.json
```diff
@@ -78,6 +78,7 @@
     "@storybook/react-webpack5": "7.5.3",
     "@testing-library/jest-dom": "^6.0.0",
     "@testing-library/react": "^14.0.0",
+    "@testing-library/user-event": "14.5.1",
```
mjmonline, line 81: Please use `^` if possible. We should try to avoid pinning because that makes it easier for apps like webapp-next using this library

tsconfig.json
gabagoo: To enable extended expects (testing-library) we can't exclude test files in the tsconfig. We followed the this guide https://github.com/testing-library/jest-dom#with-typescript.

package.json
gabagoo: But this is a devDependency so it shouldn't matter in this case? This dependency is only used by tests

src/Molecules/Input/FormattedNumber/FormattedNumber.styles.tsx
TahirMehdi: We agree that styles have to be updated, and we will definitely do it in future iteration. For now, we want to get these changes out for testing, especially since functionality tests may reveal unforeseen issues.

src/Molecules/Input/FormattedNumber/utils.ts
TahirMehdi: Updated

src/Molecules/Input/FormattedNumber/FormattedNumber.tsx
TahirMehdi: Done

src/Molecules/Input/FormattedNumber/FormattedNumber.tsx
TahirMehdi: Done

emilwestin (approved):

PR #1523, opened 2023-11-29
feat: 🎸 Add new icon for help and update selectionsCard

mic1000:


yarn.lock
```diff
@@ -0,0 +1,14563 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
davthoNordnet, line 1: Just making sure - was this intentional?

mic1000: This is auto generated so I don't know why this was in my PR

PR #1535, opened 2023-12-04
Fix 14.1.4 -> 15.0.3 bump

davthoNordnet:
image

image

image

image

PR #1536, opened 2023-12-04
fix: lookbehind regexp compatibility issue

TahirMehdi:
Lookbehind regexp is not supported by older Safari versions. 
See #1535 for more details.

### DIFF
--- a/src/Molecules/Input/FormattedNumber/FormattedInput/utils.ts
+++ b/src/Molecules/Input/FormattedNumber/FormattedInput/utils.ts
@@ -20,6 +20,18 @@ export const getMinusSign = (intl: IntlShape): string => intl.formatNumberToPart
 export const formatNumber = (value: number, intl: IntlShape, minimumFractionDigits = 0) =>
   intl.formatNumber(value, { minimumFractionDigits, maximumFractionDigits: 20 });
 
+const prependWithZero = (value: string) => {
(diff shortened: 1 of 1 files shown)


src/Molecules/Input/FormattedNumber/FormattedInput/utils.ts
```diff
@@ -38,21 +50,23 @@ const getSanitizedValue = (inputValue: string, replaceValue: string, intl: IntlS
     .replace(replaceDecimalRegExp, '.') // replace localized decimalSign
     .replace(replaceMinusSignRegExp, '-'); // replace localized minusSign
 
-  const sanitizedValue = delocalizedValue
-    .replace(/(?<=^|-)\./, '0.') // prepend 0 when integer part is missing for decimal
+  const sanitizedValue = prependWithZero(delocalizedValue) // prepend 0 when integer part is missing for decimal
     .replace(/(?<!^)-/g, replaceValue); // remove minusSign not at start of string
```
davthoNordnet, line 54: I suspect ?<!... might be an issue. Would be good to have that verified somehow.

davthoNordnet (dismissed):
Looks good. I'm a bit suspicous of the usage of '?<!...' - would be interesting to see how it behaves in a cage and older OS/browser versions!

davthoNordnet (approved):

PR #1547, opened 2023-12-11
add count and flag to inout select dropdown items

jesperwiner:


src/Molecules/Input/Select/lib/MultiSelectList/MultiSelectList.tsx
```diff
@@ -119,6 +126,7 @@ export const Option = ({
           readOnly
           visuallyFocused={!disabled && isKeyboardNavigation ? focused : false}
         />
+        {showFlag && (country ? <Flag size="m" country={country} /> : <Box p={1} />)}
```
davthoNordnet, line 129: Aren't country and showflag being used for the same purpose here - it feels like showFlag is not needed as the country prop only seems to be used when we want to display a flag

stigdanielek (approved):

src/Molecules/Input/Select/lib/MultiSelectList/MultiSelectList.tsx
jesperwiner: Need to reserve space for items without a flag to maintain column alignment

PR #1572, opened 2024-01-12
fix: the not:empty css query messes up how flex box are applies …

LifeofMichal:
…to children and grandchildren when styled-components version is updated to 16.0.0 and ahead

### DIFF
--- a/src/Atoms/Flexbox/Flexbox.tsx
+++ b/src/Atoms/Flexbox/Flexbox.tsx
@@ -72,11 +72,11 @@ const getGapStyles = (theme: Theme, gap: Props['gap']) => {
   }
 
   if (typeof gap === 'number') {
-    gapStyle = `${theme.spacing.unit(gap)}px;`;
+    gapStyle = `${theme.spacing.unit(gap)}px`;
   } else if (typeof gap === 'string') {
     gapStyle = gap;
   } else {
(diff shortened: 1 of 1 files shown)


mjmonline (dismissed):

src/Atoms/Flexbox/Flexbox.tsx
```diff
@@ -78,11 +78,8 @@ const getGapStyles = (theme: Theme, gap: Props['gap']) => {
   } else {
     gapStyle = `${theme.spacing.unit(gap.row)}px ${theme.spacing.unit(gap.column)}px;`;
   }
-  /*
-    :not(:empty) to fix issue with printing in chrome
-    https://bugs.chromium.org/p/chromium/issues/detail?id=1161709
-  */
-  return `:not(:empty) { gap: ${gapStyle}; }`;
+
+  return `gap: ${gapStyle}`;
```
mjmonline, line 82: I know I'm really annoying but this looks scary to me hehe, can we please do it the other way around? Removing the other two semicolons instead of this?

mjmonline (approved):

src/Atoms/Flexbox/Flexbox.tsx
LifeofMichal: done :D

mjmonline (approved):

PR #1578, opened 2024-01-22
feat: add support for inverted colors

jesperwiner:


stigdanielek (dismissed):

src/Molecules/BottomSheet/BottomSheet.tsx
```diff
@@ -1,35 +1,51 @@
...
+  onClose: onCloseExternal,
+  open: isOpenExternal,
   showBackdrop = true,
   title,
 }) => {
   const internalRef = useRef<HTMLDivElement>(null);
+  const isControlled = isBoolean(isOpenExternal);
+  const [isOpenInternal, setIsOpenInternal] = useState(true);
+
+  const TRANSITION_DURATION = 0.16;
```
patricksjoberg, line 27: Nitpick: Should maybe be in a const file or at least outside of the component scope?

src/Molecules/BottomSheet/BottomSheet.tsx
```diff
@@ -1,35 +1,51 @@
...
-  open,
+  height,
+  invertedColors,
+  onClose: onCloseExternal,
+  open: isOpenExternal,
   showBackdrop = true,
   title,
 }) => {
   const internalRef = useRef<HTMLDivElement>(null);
+  const isControlled = isBoolean(isOpenExternal);
```
patricksjoberg, line 24: Do you want to convert falsy and truthy values to booleans here?
If so, I think you should use Boolean(isOpenExternal) because this only checks if it is a boolean type. So if somebody sets isOpenExternal to false isControlled would be true because false is a boolean type.

src/Molecules/BottomSheet/BottomSheet.tsx
```diff
@@ -1,35 +1,51 @@
...
+  const [isOpenInternal, setIsOpenInternal] = useState(true);
+
+  const TRANSITION_DURATION = 0.16;
+
+  const onClose = () => {
+    setIsOpenInternal(false);
+
+    if (typeof onCloseExternal === 'function') {
```
patricksjoberg, line 32: Must we have this check? Can't we just use `onCloseExternal?.()` ?

src/Molecules/Tooltip/Tooltip.tsx
```diff
@@ -80,14 +68,43 @@ export const Tooltip: FC<Props> = (props) => {
...
+
+  const handleOpenBottomSheet = (e: React.MouseEvent) => {
+    if (controlledIsOpen === undefined) {
+      setBottomSheetOpen(true);
+      e.stopPropagation();
+    }
+  };
+
+  const handleCloseBottomSheet = () => {
+    if (typeof onBottomSheetClose === 'function') {
```
patricksjoberg, line 81: same here. `onBottomSheetClose?.()`

patricksjoberg (changes requested):
Nice job! Maybe hade one thing that is blocking.

src/Molecules/BottomSheet/BottomSheet.tsx
jesperwiner: its copy paste from another one like this

src/Molecules/BottomSheet/BottomSheet.tsx
jesperwiner: not the same thing, then you can send an object and it will pass the check and then runtime error fail when trying to run it

src/Molecules/Tooltip/Tooltip.tsx
jesperwiner: same as above

patricksjoberg (dismissed):

Garima-Kaila (approved):

stigdanielek (approved):

PR #1594, opened 2024-02-19
fix: set default knob according to distance to min and max

jesperwiner:


src/Molecules/RangeSlider/RangeSlider.tsx
```diff
@@ -17,6 +17,23 @@ const Container = styled.div<InternalProps>`
...
 
+const getDefaultKnob = (
+  low: number | undefined,
+  high: number | undefined,
+  min: number,
+  max: number,
+) => {
+  if (low && high && low === high) {
+    const distanceToMin = low - min;
+    const distanceToMax = max - high;
```
mikaelmalmqvistnordnet, line 28: you might wanna use the absolute value of the diff instead to avoid weird negative values
```suggestion
    const distanceToMin = Math.abs(low - min);
    const distanceToMax = Math.abs(max - high);
```

src/Molecules/RangeSlider/RangeSlider.tsx
```diff
@@ -17,6 +17,23 @@ const Container = styled.div<InternalProps>`
   margin: 0 ${(p) => `${getKnobSize(p.$variant) / 2}px`};
 `;
 
+const getDefaultKnob = (
+  low: number | undefined,
+  high: number | undefined,
+  min: number,
+  max: number,
+) => {
+  if (low && high && low === high) {
```
mikaelmalmqvistnordnet, line 26: you won't enter this if-statement if `low` or `high` is 0. Is that intended?

src/Molecules/RangeSlider/RangeSlider.tsx
```diff
@@ -17,6 +17,23 @@ const Container = styled.div<InternalProps>`
   margin: 0 ${(p) => `${getKnobSize(p.$variant) / 2}px`};
 `;
 
+const getDefaultKnob = (
+  low: number | undefined,
+  high: number | undefined,
```
mikaelmalmqvistnordnet, line 22: Looking at the types in `Props` (src/Molecules/RangeSlider/RangeSlider.tsx) that these are based on, they are optional so you could simplify this with just the `?`-operator. OR if you wanna be really specific and, you can pick the correct type from `Props` instead like `Props['defaultHighValue']`

src/Molecules/RangeSlider/RangeSlider.tsx
```diff
@@ -133,7 +147,7 @@ const RangeSlider: Component = ({
       $disabled={disabled}
       $sliderColor={sliderColor}
       $variant={variant}
-      onClick={handleTrackClick}
+      onMouseDown={handleTrackClick}
```
mikaelmalmqvistnordnet, line 150: add a comment here explaining what you explained to me, so that future devs understand why we don't use onClick here and don't change back by mistake

mikaelmalmqvistnordnet (approved):

PR #1600, opened 2024-02-28
New toast message component

mic1000:
https://jira.prod.nordnet.se/browse/DISCO-11350

Toasts inform users of a process that an app has performed or will perform. They appear temporarily, towards the bottom of the screen. They shouldn’t interrupt the user experience, and they don’t require user input to disappear.

Toasts provide brief messages about app processes at the bottom of the screen.
Only one toast may be displayed at a time.
Toasts appear temporarily, and disappear on their own without requiring user to dismiss it.
A toast can contain a single action.

### DIFF
--- a/src/Molecules/ToastMessage/ToastMessage.stories.tsx
+++ b/src/Molecules/ToastMessage/ToastMessage.stories.tsx
@@ -1,0 +1,217 @@
+import React, { useState, useRef } from 'react';
+import { action } from '@storybook/addon-actions';
+
+import ToastMessage from './ToastMessage';
+import { Display } from '../../common/Display';
+import { Icon, Button, PageWrapper, Flexbox, Box, Badge } from '../..';
+
+export default {
+  title: 'Molecules / ToastMessage ',
+  parameters: {
+    component: ToastMessage,
+  },
+};
+
+export const DifferentVariantToastMessages = {
+  render: () => (
+    <>
+      <Display
+        items={[
+          {
+            title: 'primary',
+            component: (
+              <ToastMessage
+                label="Link copied to clipboard"
+                icon={<Icon.CheckCircleFill24 color={(t) => t.colorTokens.positive.icon_default} />}
+                isVisible
+                linkButton={
+                  <Button
+                    onClick={action('clicked')}
+                    variant="neutral"
+                    color={(t) => t.colorTokens.action.text_focus}
+                  >
+                    View
+                  </Button>
+                }
+              />
+            ),
+          },
+        ]}
+      />
+    </>
+  ),
+
+  name: 'Different variants',
+};
+export const HoverUsage = {
+  render: () => {
+    const Example = () => {
+      const [visible, setVisible] = useState<boolean>(false);
+
+      return (
+        <PageWrapper background={(t) => t.colorTokens.neutral.background_brand_white}>
+          <Flexbox>
+            <Box onMouseEnter={() => setVisible(true)} onMouseLeave={() => setVisible(false)} p={4}>
+              <Badge.Status variant="create" badgeSize="xl" label="Hover me" />
+            </Box>
+          </Flexbox>
+          <Flexbox container gap={10}>
+            <ToastMessage
+              label="Link copied to clipboard"
+              icon={<Icon.CheckCircleFill24 color={(t) => t.colorTokens.positive.icon_default} />}
+              isVisible={visible}
+              linkButton={
+                <Button
+                  onClick={action('clicked')}
+                  variant="neutral"
+                  color={(t) => t.colorTokens.action.text_focus}
+                >
+                  View
+                </Button>
+              }
+            />
+          </Flexbox>
+        </PageWrapper>
+      );
+    };
+    return <Example />;
+  },
+  name: 'With Hover',
+};
+export const WithLabelUsage = {
+  render: () => {
+    const Example = () => {
+      const [visible, setVisible] = useState<boolean>(false);
+
+      return (
+        <PageWrapper background={(t) => t.colorTokens.neutral.background_brand_white}>
+          <Flexbox>
+            <Button
+              variant="primary"
+              onMouseEnter={() => setVisible(true)}
+              onMouseLeave={() => setVisible(false)}
+            >
+              Hover me
+            </Button>
+          </Flexbox>
+          <Flexbox container gap={10}>
+            <ToastMessage label="Saved to watchlist" isVisible={visible} />
+          </Flexbox>
+        </PageWrapper>
+      );
+    };
+    return <Example />;
+  },
+  name: 'With label',
+};
+export const WithLabelUsageLink = {
+  render: () => {
+    const Example = () => {
+      const [visible, setVisible] = useState<boolean>(false);
+
+      return (
+        <PageWrapper background={(t) => t.colorTokens.neutral.background_brand_white}>
+          <Flexbox>
+            <Button
+              variant="primary"
+              onMouseEnter={() => setVisible(true)}
+              onMouseLeave={() => setVisible(false)}
+            >
+              Hover me
+            </Button>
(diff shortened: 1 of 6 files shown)


src/Molecules/ToastMessage/ToastMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+          setVisibility(false);
+        }
+      }, timeout);
+      return () => clearTimeout(timer);
+    }, [timeout, visible, mouseEnter]);
+
+    return (
+      <AnimatePresence mode="wait">
+        {visible && (
+          <motion.div key="toast" variants={variants} initial="initial" animate="enter" exit="exit">
```
jesperwiner, line 59: think you need to add a unique id to the key too, in case we ever end up with multiple on same page

src/Molecules/ToastMessage/ToastMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+        opacity: 1,
+        transition: {
+          duration,
+          ease: 'easeInOut',
+        },
+      },
+      exit: {
+        opacity: 0,
+        transition: { duration },
+      },
```
jesperwiner, line 28: duration can be set outside the variants since its same 

transition={{ duration: 0.35 }}

src/Molecules/ToastMessage/ToastMessage.styled.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const BoxContainer = styled(Box)`
+  background-color: ${({ theme }) => theme.colorTokens.neutral.background_focus};
+  padding: ${(p) => p.theme.spacing.unit(3)}px;
+  border-radius: ${(p) => p.theme.borderRadius8};
+  width: fit-content;
+  position: fixed;
+  right: 20px;
+  bottom: 20px;
+  z-index: 3;
```
jesperwiner, line 13: might need to be like 
z-index: ${({ theme }) => theme.zIndex.overlayInModal;
to be ontopest

src/Molecules/ToastMessage/ToastMessage.tsx
mic1000: Good point

src/Molecules/ToastMessage/ToastMessage.stories.tsx
```diff
@@ -0,0 +1,208 @@
...
+            ),
+          },
+        ]}
+      />
+    </>
+  ),
+
+  name: 'Different variants',
+};
+export const HoverUsage = {
```
jesperwiner, line 46: enter :)

src/Molecules/ToastMessage/ToastMessage.styled.ts
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components';
+
+import { Box, Typography } from '../..';
+
+export const BoxContainer = styled(Box)`
+  background-color: ${({ theme }) => theme.colorTokens.neutral.background_focus};
+  padding: ${(p) => p.theme.spacing.unit(3)}px;
+  border-radius: ${(p) => p.theme.borderRadius8};
```
mikaelmalmqvistnordnet, line 8: use destructuring on all places here for consistency
```suggestion
  background-color: ${({ theme }) => theme.colorTokens.neutral.background_focus};
  padding: ${({ theme }) => theme.spacing.unit(3)}px;
  border-radius: ${({ theme }) => theme.borderRadius8};
```

src/Molecules/ToastMessage/ToastMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { AnimatePresence, motion } from 'framer-motion';
+
+import { Flexbox } from '../..';
+import { BoxContainer, StyledTypography } from './ToastMessage.styled';
+import { Props } from './ToastMessage.types';
+
+const ToastMessages: React.FC<Props> = React.forwardRef(
+  ({ label, icon, linkButton, isVisible, timeout = 6000 }, ref) => {
+    const [visible, setVisibility] = useState<boolean>(isVisible || false);
+    const [mouseEnter, setMouseEnter] = useState<boolean>(false);
```
mikaelmalmqvistnordnet, line 11: no need to explicitly type the `useState` here as you get the type implicitly by the init value here
```suggestion
    const [visible, setVisibility] = useState(isVisible || false);
    const [mouseEnter, setMouseEnter] = useState(false);
```

src/Molecules/ToastMessage/ToastMessage.tsx
mikaelmalmqvistnordnet, line 28: Since `variants` is not of a primitive type, it will get re-evaluated each time this component re-renders, which is not ideal. An easy fix would be to move it out of the component :)

src/Molecules/ToastMessage/ToastMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+      }
+
+      if (mouseEnter) {
+        setVisibility(true);
+      }
+    }, [isVisible, ref, mouseEnter]);
+
+    useEffect(() => {
+      const timer = setTimeout(() => {
+        if (visible === true && !mouseEnter) {
```
mikaelmalmqvistnordnet, line 49: no need to explicitly compare with `true` since `visible` is a boolean
```suggestion
        if (visible && !mouseEnter) {
```

src/Molecules/ToastMessage/ToastMessage.types.ts
```diff
@@ -0,0 +1,11 @@
+import { ReactNode, Ref } from 'react';
+
+export type Props = {
+  label: string | ReactNode;
+  icon?: ReactNode;
+  linkButton?: ReactNode;
+  isVisible?: boolean;
+  timeout?: number;
+  setIsVisible?: (set: any) => void;
```
mikaelmalmqvistnordnet, line 9: fix the type here?
```suggestion
  setIsVisible?: (visible: boolean) => void;
```

src/Molecules/ToastMessage/ToastMessage.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { useEffect, useImperativeHandle, useState } from 'react';
+import { AnimatePresence, motion } from 'framer-motion';
+
+import { Flexbox } from '../..';
+import { BoxContainer, StyledTypography } from './ToastMessage.styled';
+import { Props } from './ToastMessage.types';
+
+const ToastMessages: React.FC<Props> = React.forwardRef(
+  ({ label, icon, linkButton, isVisible, timeout = 6000 }, ref) => {
```
mikaelmalmqvistnordnet, line 9: Since `isVisible` is optional in your types, you can default it to `false` to make sure you're always dealing with a boolean

```suggestion
  ({ label, icon, linkButton, isVisible = false, timeout = 6000 }, ref) => {
```

src/Molecules/ToastMessage/ToastMessage.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    useImperativeHandle(ref, () => ({
+      setVisible: () => {
+        setVisibility(!visible);
+      },
+    }));
+
+    useEffect(() => {
+      if (typeof ref === 'undefined' || ref === null) {
+        setVisibility(isVisible || false);
```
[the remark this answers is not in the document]
mic1000: setVisibility should not be used if ref is available because ref will dictate the control

src/Molecules/ToastMessage/ToastMessage.tsx
mic1000: I can move it out NP

mikaelmalmqvistnordnet (approved):
Looks great :D

jesperwiner (approved):

